PR #7, opened 2013-11-05
ubuntu install script

mk-pmb:
Hi, this script checks the most important deps, installs them if needed, and goes on to compile Drawpile.


mk-pmb:
I was considering some minor improvements to my old code for this, but found it was removed in f037c029f68eff14c1082a486cfd2f6c75fb4f13 anyway so it doesn't matter. Writing this just so I can remember if I ever get the same idea again.

Issue #100, opened 2014-07-10
Tablet pen eraser does not revert to pencil/brush when turned, nor restore settings


GreenReaper:
If the tablet pen is turned over to use the eraser, the eraser is correctly selected, but when the user turns the pen back, the tool remains an eraser. This is very confusing to users, especially with #99.

Not only this, but when the user re-selects the pen or brush, they find that the settings are different. The settings should be restored to those present at the point that it turned into an eraser, e.g. if it is a 1px brush with 50% opacity to start with, it should turn back to that.

PR #196, opened 2015-05-23
Adding possible workaround for problems with Qt QEvent::Tablet(Enter|…

mlichter:
…Leave)Proximity events.

I still found behaviour similar Issue #100 under Linux after recompiling Drawpile against Qt5.4.1. It seems that after clicking something in the menu, like you described in issue #100, Drawpile sees Enter and Leave events two times, instead of the Enter Leave combination under prior Qt Versions.
When this happens the actual tool gets overwritten by the eraser tool.

I tried to workaround that by tracking if the Eraser was selected that way and only changing state if it was/wasn't selected before.
For Qt versions prior to 5.4 the switching does work if one only reacts to the EnterProximity event from eraser and pen. With the minor drawback that the eraser tool stays active until you bring the pen tip in proximity of the tablet again.

### DIFF
--- a/src/client/docks/toolsettingsdock.cpp
+++ b/src/client/docks/toolsettingsdock.cpp
@@ -34,7 +34,7 @@
 namespace docks {
 
 ToolSettings::ToolSettings(QWidget *parent)
-	: QDockWidget(parent), _currentQuickslot(0), _eraserOverride(0)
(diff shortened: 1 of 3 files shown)


callaa:
Thanks! I tested this on Linux and OSX and it seems to work.

PR #217, opened 2015-07-12
I18n de

mlichter:
German translation for Drawpile

The translation should be pretty complete at this moment. There is some fuzz about the translations for the client/server part of Drawpile, but it looks ok for me. Corrections and feedback would be highly appreciated.


callaa:
Thanks! Just in time for 1.0.0 release!

Issue #179, opened 2015-02-26
Feature: real stabilizer for classic brushes


blurymind:
What it is:
Its a value that slows down your stroke. This makes it much easier to draw precise, smooth strokes.

Some people call it "lazy mouse" - zbrush.
In blender it is called "smooth stroke"
In krita it is called a "stabilizer" - they even have a number of variants of this same feature.
https://www.youtube.com/watch?v=YpXT_Mx_u6k

In inkscape, it is the "Mass" value of the calligraphy pen tool
https://inkscape.org/en/doc/calligraphy/tutorial-calligraphy.html
"In physics, mass is what causes inertia; the higher the mass of the Inkscape calligraphy tool, the more it lags behind your mouse pointer and the more it smoothes out sharp turns and quick jerks in your stroke. By default this value is quite small (2) so that the tool is fast and responsive, but you can increase mass to get slower and smoother pen."

and so on.
I literally cant ink without this feature. :)

callaa:
Right now Drawpile has a simple sliding average smoother, which I think is equivalent to Krita's "Basic Smoothing". (Or at least seems to be, I haven't checked the source code yet.)
Krita's "Weighted Smoothing" seems to be the feature you describe. I'll take a look at how it works and see if a similar feature will be easy to implement.

blurymind:
Thank you! Looking forward to it :)

It would be very convenient if it can be assigned on per individual brush basis. So my inking pen for example would be set to "weighted smoothing" of 1 for example

LionsPhil:
I've done some digging through the (massive) Krita codebase and from I can tell their smoothing is mostly implemented in [kis_tool_freehand_helper.cpp](https://projects.kde.org/projects/calligra/repository/revisions/master/entry/krita/ui/tool/kis_tool_freehand_helper.cpp) (with parameters set [with the brush tool](https://projects.kde.org/projects/calligra/repository/revisions/master/entry/krita/plugins/tools/defaulttools/kis_tool_brush.cc)).

[...]

PR #219, opened 2015-07-12
Remove starting lag from smoothed strokes

LionsPhil:
The current smoother causes a pretty nasty positional lag to the start of strokes, which makes it hard to accurately position them to connect with other strokes, e.g. drawing a set of lines radiating from a central point.

Have smoothing effectively start at zero and ramp up as more points become available so strokes start where the cursor first touches down.

Partially relates to #179, but the smoother is still just a plain unweighted average.

(I think in future it might be worth looking at having the smoother buffer up some move events so it can still use as many points as possible, but cause *temporal* lag instead, e.g. when you have a full buffer suddenly canvasview can drain that many moves from the smoother, after which it is back to one point added, one (lagged) move received.)

### DIFF
--- a/src/client/scene/canvasview.cpp
+++ b/src/client/scene/canvasview.cpp
@@ -437,8 +437,7 @@ void CanvasView::onPenDown(const paintcore::Point &p, bool right)
 		} else {
 			if(_smoothing>0 && _current_tool->allowSmoothing())
 				_smoother.addPoint(p);
-			else
-				_current_tool->begin(p, right, _zoom);
+			_current_tool->begin(p, right, _zoom);
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


callaa:
Thanks, I'll take a look at this. I just released version 1.0 and started a new stable branch. Since this does not affect the protocol and is a small change, I'll probably merge it to *stable-1.x* so it can be part of the next (1.0.1) release.

Buffering hovering pointer events is a good idea. MyPaint does that too, although for brush dynamics rather than stabilizing, I think.

LionsPhil:
I'll rebase this onto 1.0, with another tweak commit or two. Doing them separately to get the effect of each change in isolation, but I can squash them down if you'd rather since they're individually tiny.

callaa:
No need to rebase or squash since there will be a merge commit anyway.

LionsPhil:
OK, that's the last of them for now, then. Afraid there's a little bit of thrash where I reintroduced latestPoint() after removing it earlier.

PR #280, opened 2015-11-21
Dockerised Drawpile Server definition

jalev:
One of the things which I found painful during the setup of a drawpile server was the amount of dependencies I needed to install to get a dedicated `drawpile-srv` up and running. 

To make things more accessible, I've created a [Docker](https://www.docker.com/) spec for `drawpile-srv`. This will reduce the amount of setup a developer/potential hoster needs to something like `docker run --daemonize -p 8080:8080 drawpile` (it's trivial to host the build spec on [Docker Hub](https://hub.docker.com/) and have it build the container automatically, available for people to host as they want).

Below is an example of me running a drawpile server on a hastily spun up AWS vm. Total time between me starting this VM and getting drawpile up and running: <5mins.

Drawpile on Docker containers

### DIFF
--- a/server/docker/Dockerfile
+++ b/server/docker/Dockerfile
@@ -1,0 +1,33 @@
+FROM ubuntu:15.04
+RUN apt-get update \
+ && apt-get install \
+    wget \
+    cmake \
+    make \
+    qtbase5-dev \
+    libkf5archive-dev \
+    g++ \
+    qtm \
+    qtmultimedia5-dev \
+    libqt5svg5-dev \
+    libgif-dev \
+    libkf5dnssd-dev \
+    -y \
+ && rm -rf /var/lib/apt/lists/* /var/cache/apt/archives/*
+
+ENV VERSION 1.0.2
+
+RUN wget http://drawpile.net/files/src/drawpile-$VERSION.tar.gz \
+ && tar -xf drawpile-$VERSION.tar.gz \
+ && cd drawpile-$VERSION \
+ && mkdir build \
+ && cd build \
+ && cmake .. \
+ && make \
+ && make install
+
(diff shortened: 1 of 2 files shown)


callaa:
Thanks!
If you add `-DCLIENT=OFF` to the cmake line, you can remove qtm, qtmultimedia, qtsvg, libgif and kdnssd libraries. Unfortunately Ubuntu doesn't ship a GUI-less version of Qt development packages, so it doesn't seem to be possible to avoid pulling in X and other dependencies. (Except perhaps by having separate build and deployment images.)
Also, it might be better to put the Dockerfile into a directory of its own to avoid uploading the entire source directory as context, since the source is fetched with wget anyway.

Any specific reason why you use 8080 as the port rather than the default one?

Last, the `--daemonize` argument doesn't seem to work with my docker version.

jalev:
I'm always worried about file-sizes when I have to Dockerize something with X as a dependency. Unless headless Qt dev packages exist, the docker container will always clock in at ~600MB, which is aggravating. However- this constraint only exists because the Dockerfile is compiling from source. We can absolutely cut this down to a fraction with a 1.0.2 deb package, which is what I'm looking at doing next. 

The Dockerfile is stored in its own folder within within `server` atm. I figured that would be appropriate since it's only giving you a server, and not containing any source. 
 
Also, woops. I meant `--detatch`, not `--daemonize`! Thanks for catching that (I don't usually use the expanded command flags). 

Also, 8080 is just a number I use a lot when doing Docker. I'm not sure what the default port `drawpile-srv` uses when you first run it, so I just took standard webby ones to be safe. I'll take a look into that too, since you have to enter the ports when connecting to a Drawpile server atm, and that's terrible.

PR #329, opened 2016-06-01
actually sort sessions by UserRole rather than DisplayRole

timo:
This is what sessionlistingmodel.cpp claims is already the case,
but the default value is actually DisplayRole.


callaa:
Thanks!

PR #507, opened 2017-08-18
Make it pass desktop-file-validate

probonopd:
`desktop-file-validate` currently says: `drawpile.desktop: error: value "image/openraster;image/png;image/jpeg;application/x-drawpile-recording;x-scheme-handler/drawpile" for string list key "MimeType" in group "Desktop Entry" does not have a semicolon (';') as trailing character`


callaa:
Thank you! Sorry it took so long for me to take a look at this.

PR #612, opened 2018-05-06
mention Dockerfile in README.md

vn971:
Dockerfile helps a lot when building on Debian/Ubuntu. The `apt-get install` line is pretty re-usable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Client specific dependencies:
 * GIFLIB (optional: animated GIF export)
 * MiniUPnP (optional: automatic port forwarding setup)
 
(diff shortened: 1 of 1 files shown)


vn971:
@callaa Dockerfile helped me getting the exact dependencies a lot. It's not easy to understand which of the Qt you need - but the Dockerfile must verifiably tell just that.

vn971:
// That being said, I'm not a real fan of neither AppImage nor Docker myself, so I was happy I'm just able to compile & run Drawpile as a normal app. It's all opinions, anyway - no truth.

PR #629, opened 2018-06-20
Update Chinese translation

Sansui233:
- Translate session data dialog part
- Improve translation to make it more in line with Chinese habits
- Modify some keywords' translations referring to  the similar software [Mr.Paint](https://github.com/liuyanghejerry/painttyWidget) which is developed by Chinese, such as from "连接" to "房间" about keyword "sessions" .
- Many small supplements


callaa:
Thank you!

Issue #801, opened 2019-09-20
master fails to build, complaining about QDeadlineTimer


vn971:
Hi. I'm trying to build Drawpile (master branch).
Cmake runs successfully:
```
cmake .. -DCLIENT=off -DSERVERGUI=off -DINITSYS=""
-- No build type selected, default to Release
-- Use -DCMAKE_BUILD_TYPE=debug to enable debugging
-- Build type: Release
CMake Warning at src/thinsrv/CMakeLists.txt:74 (message):
  Libmicrohttpd not found: web-admin not enabled


-- CXX flags (release): 
-- Configuring done
-- Generating done
-- Build files have been written to: /home/pavel/Drawpile/build
```

But when trying to run `make`, I get this error:
```
[ 48%] Building CXX object src/libserver/CMakeFiles/dpserver.dir/session.cpp.o
In file included from /home/pavel/Drawpile/src/libserver/session.cpp:25:0:
/home/pavel/Drawpile/src/libserver/announcements.h:26:26: fatal error: QDeadlineTimer: No such file or directory
 #include <QDeadlineTimer>
                          ^
compilation terminated.
src/libserver/CMakeFiles/dpserver.dir/build.make:110: recipe for target 'src/libserver/CMakeFiles/dpserver.dir/session.cpp.o' failed
make[2]: *** [src/libserver/CMakeFiles/dpserver.dir/session.cpp.o] Error 1
CMakeFiles/Makefile2:199: recipe for target 'src/libserver/CMakeFiles/dpserver.dir/all' failed
make[1]: *** [src/libserver/CMakeFiles/dpserver.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

Might there be something wrong with how I compile Drawpile?
Thanks

Wade821:
No, you simply have an outdated version of QT causing the build error. You can either update QT if there's a newer download available from your repo, compile QT from source, modify the source code accordingly in order to make it compile using your existing installation, or [try running my script ](https://github.com/Wade821/PiDrawpile) to compile the server from source and run the necessary code changes automatically. The script itself has sections that you can use for notes on required changes to the source code that you should be able to find if you decide to do it manually yourself. I'd help but I'm about to leave for work for the day. 

That being the case, these are the commands needed to get started if compiling Drawpile from Source on Linux. :smiley: The script will be there to walk you through the rest of the necessary steps. Improvements to come on the script implementation later, working on a major re-write. 

```
[...]

PR #851, opened 2019-12-27
add support for Qt versions lower than 5.8

vn971:
related to GH-801.

### DIFF
--- a/src/libserver/loginhandler.cpp
+++ b/src/libserver/loginhandler.cpp
@@ -305,7 +305,9 @@ void LoginHandler::handleIdentMessage(const protocol::ServerCommand &cmd)
 				return;
 			}
 		}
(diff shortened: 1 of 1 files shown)


callaa:
I think a better approach would be to add this to the top of the file (but after the Qt includes):

    #ifndef Q_FALLTHROUGH
    #define Q_FALLTHROUGH() (void)0
    #endif

The fallthough macro expands to a compiler specific fallthrough tag, or (void)0 for compilers that don't support it. It doesn't actually do anything, except suppress a warning act remind the programmer that the fallthrough is intentional.

vn971:
Sure, I don't mind this other approach as well. I've pushed the change. (It compiles successfully.)

callaa:
Thanks.

PR #855, opened 2020-01-08
Fix favicon download not saving to conf

notfood:
connect is an asynchronous operation, the code will run after the image url is loaded. It's not saving the changes made by setFavicon into drawpile.conf.


callaa:
Thanks!

PR #897, opened 2020-05-14
Fix brush size adjust shortcuts not updating brush size correctly

taiyu-len:
currently on my computer, updating the brush(or whichever tool it is) size via the [ or ] keys requires 2 key presses to show any effect, and will change brush size by 2 when it does.

The change is to compare the integer component of m_quickAdjust1 directly to avoid any floating point comparisions, which seem to be the cause of the issue.

This also makes all the quickAdjust1 overrides consistent, before they differed in whether the comparison was `>= 1.0` or `> 1.0`.


taiyu-len:
made a quick change, from `qAbs(int(i)) >= 1` to just `int(i)` since what its really doing is that brush size should change when the integer component of m_quickAdjust1 is non-zero.

callaa:
Thank you!

PR #898, opened 2020-05-14
Fix being unable to move entire canvas

taiyu-len:
selection range is exclusive, so width and height of selection are one bigger then the canvas size.

adds 1 to width and height when calculating the sanity check

I havent tested it in a online session yet, so not sure if this is something that requires all clients/server to agree upon or not to avoid inconsistencies, which may be an issue.


callaa:
Thank you! Selection operations are local only, so this should work just as well in online mode.

PR #899, opened 2020-05-15
Fix closed selections being initially unclosed

### LINKED ISSUE #881: Trouble pasting images bigger than canvas
Whenever you paste images bigger than the canvas in any axis, you can't use the handlers to resize it. When you attempt to resize the selection gets rotated -90 degrees, trimmed to the size of the canvas and your image loses its ratio, making it very hard to work with.

Workaround:
Use selection tool's fit to screen option to resize the image without touching the handlers.

Relevant commits: b3385b0bd15aa42ba6f5f5037e49fc8d0e696f55 2d18cbfa121646826a1a051c0b6d4acac6554357 42f4ef6925743405a4d99fd1093082eb5fea2bd2

taiyu-len:
Fixes #881 
Currently the selections created by pasting and "Select All" start unclosed.
Once moved/clicked the `Selection::closeShape(clipRect)` function is called, which distorts the image.

These selections which are complete from the start should be closed on creation, 
This fix adds back the `Selection::closeShape()` function that does not clip, and uses that for the above commands on creation of the Selection.

There may be other such selections that have this behavior that may need to be fixed in the same way.

### DIFF
--- a/src/libclient/canvas/selection.cpp
+++ b/src/libclient/canvas/selection.cpp
@@ -306,6 +306,15 @@ void Selection::addPointToShape(const QPointF &point)
 	}
 }
 
+void Selection::closeShape()
+{
+	if(!m_closedPolygon) {
+		m_closedPolygon = true;
+		saveShape();
+		emit closed();
+	}
+}
+
 bool Selection::closeShape(const QRectF &clipRect)
 {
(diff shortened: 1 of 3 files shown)


callaa:
Thank you!

PR #905, opened 2020-05-27
Color picker cursor fix

### LINKED ISSUE #839: Color picker cursor sometimes stays after use
No idea if this has been addressed in the pending build, but I thought I'd put this here for visibility.

When using the shortcut to pick colors, sometimes the cursor stays as the eyedropper, even though it is actually in brush mode. This usually corrects itself after using color picker again, though if needed, whenI get this issue, I can see if hitting b for the brush tool shortcut also makes the cursor go back to normal.

MataMercer:
Fixes #839 

The cursor would still have color picker cursor after color picking and moving at the same time. This commit resets the cursor when the pen is lifted or no longer in color picking mode. 

This is my first pull request, so it's pretty small and I'm trying figure things out.

### DIFF
--- a/src/desktop/scene/canvasview.cpp
+++ b/src/desktop/scene/canvasview.cpp
@@ -491,6 +491,8 @@ void CanvasView::onPenUp()
 		emit penUp();
 
 	m_penmode = PenMode::Normal;
+	resetCursor();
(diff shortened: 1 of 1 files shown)


callaa:
Thank you!

PR #918, opened 2020-07-18
Set initial format for annotation content

taiyu-len:
Current behavior is that when using the annotation tool for the first time, creating a new annotation, the contents of the `ui->content` do not use the expected default format as shown in the formatting settings.

This PR will explicitly set the right font family, size, and color to the expected values.
This PR also adds the `QPainterPath` header to 2 files in order to compile correctly.


taiyu-len:
rebased on top of https://github.com/drawpile/Drawpile/commit/b46115160c5587aae9306e57d4223906e74ed512

callaa:
Thank you!

PR #921, opened 2020-08-06
Add HSV Blend Modes: Hue, Saturation, Luminosity and Color

askmeaboutlo0m:
I recently learned of a painting technique where you paint everything in grayscale first and then add the color later. I quite like that method, since it lets me split up the cognitive load of figuring out the shapes and the colors separately. However, Drawpile doesn't support a "Color" blend mode, which is kinda necessary to work with this technique.

This pull request implements that Color blend mode together with the related HSL modes Hue, Saturation and Luminosity. Those are all supported by the SVG spec as well, so I think it's reasonable to add them all.

This tragedy comes in three parts:

1. Making this build on my system. I'm on Arch Linux, so I think my library versions are newer than they are for someone on a reasonable distribution. I've had to `#include <QPainterPath>` in some places and add compatibility for the `MHD_Result` enum introduced in libmicrohttpd 0.9.71.

2. Adding the actual blend modes. Implements the HSL blending and changes around some of the rasterop functions to avoid duplicating the pixel composite functions. It's still template functions generated at compile time though, so no added dynamicity that could slow this hot code down.

3. Make the brush preview more useful for these modes. Normally, the rainbow dabs in the preview only vary in their hue, but for these HSL modes, they vary their saturation and value as well.

I've provided details in the individual commit messages as well.

### DIFF
--- a/src/desktop/toolwidgets/brushsettings.cpp
+++ b/src/desktop/toolwidgets/brushsettings.cpp
@@ -127,7 +127,14 @@ struct BrushSettings::Private {
 
 	const input::Preset *currentPreset()
 	{
-		return presetFor(currentTool());
+		const input::Preset *p = presetFor(currentTool());
+		if(!p) {
+			if(presetModel->rowCount()>0) {
+				p = presetModel->at(0);
+				currentTool().inputPresetId = p->id;
+			}
+		}
+		return p;
 	}
 
 	const input::Preset *presetFor(const ToolSlot &tool)
--- a/src/libclient/canvas/inputpresetmodel.cpp
+++ b/src/libclient/canvas/inputpresetmodel.cpp
@@ -139,10 +139,7 @@ void PresetModel::update(int index, const Preset &preset)
 void PresetModel::restoreSettings()
 {
 	beginResetModel();
-	m_presets.clear();
-
 	QSettings cfg;
-
 	const int size = cfg.beginReadArray("inputpresets");
 	m_presets.resize(size);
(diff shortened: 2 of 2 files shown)


notfood:
Tested your patch, it works fine. My suggestion is to drop the HSL prefix.

askmeaboutlo0m:
Would be easy enough, but I looked at several other drawing programs and their naming seemed to be inconsistent enough that I added in the prefix for clarity. For example, whatever version of Paint Tool SAI I have calls its Lighten blend mode Luminosity. Clip Studio calls the Luminosity mode Brightness. And GIMP has a HSL Color layer, but HSV Hue/Saturation/Value. And of course Drawpile itself has a Recolor brush mode, so removing the prefix from Color feels like inviting confusion between them.

I dunno, that was my thought process for having the prefix anyway. If there's reason to leave it off instead, that's a trivial change.

callaa:
Thank you!
This is a protocol changing feature, so I'll have to hold until version 2.2 to merge this.

askmeaboutlo0m:
I don't think it's actually a change on the protocol? It's just a few new values added to the set of blend modes, which servers just pass through as-is and clients will ignore if they don't understand it.

Unless that behavior is already enough for 2.2, in which case, alright, makes sense.

callaa:
It's a backward compatible change, but can still cause problems if some users in the session are using an older version. Old clients will interpret the new modes as the default mode. For layers this is not catastrophic (at least until a layer is merged,) but for brushes this will cause the canvas to have very different pixel values.

PR #923, opened 2020-08-09
Resize canvas with given selection

notfood:
- Allows the user to select an area and then follow with `Edit > Resize Canvas...`, the selection will be carried over to the widget. Makes resizing easier and accurate.
- Clears the selection before resize to prevent weird undo.
- This PR won't compile without the QPainterPath include fixes.


callaa:
Thank you!

PR #939, opened 2020-12-10
Add Screen layer and brush blend mode

askmeaboutlo0m:
The counterpart to the Multiply blend mode. Drawpile only had a Divide
blend mode, which is not useful to say the least.

This introduces a slight breaking change: Drawpile used to save ORA
files with its divide blend mode saved as the "screen" SVG blend mode.
That was wrong though, which is blindingly obvious if you open a file
like that in GIMP and see the result being totally different.

That means any prior Drawpile ORA files that had a Divide layer in them
(all zero of them probably, since Divide is just so utterly useless)
will now load as Screen layers instead. Any ORA files saved by other
programs using Screen layers will now load correctly.


callaa:
Thank you! As with the other pull request, I'll have to hold it until version 2.2. However, I can cherry pick the fix for the SVG mode name, since that has no effect on the network protocol.

askmeaboutlo0m:
Supersesded by #998.

PR #941, opened 2020-12-12
Mark layers dirty after removing previews

### LINKED ISSUE #940: Canceling selection transformation after another user performs an undo doesn't restore the original view
Steps to reproduce:

1. User A draws a brush stroke.

2. User B selects something and moves the selection far enough away to not touch the original tiles anymore.

3. User A undoes their stroke.

4. User B hits escape to cancel their transformation.

Expected result: User B's selection preview disappears and the layer is restored to the way it was before they started moving their selection.

Actual result: the layer *appears* to not be restored, making it look like the user just lost their drawing, usually causing some panic to ensue. Hiding and re-showing the layer, scribbling over it or doing something else to dirty that region will cause the drawing to reappear.

I have a fix for this that seems to work, but may be a bit heavy-handed. A pull request is forthcoming.

askmeaboutlo0m:
Fixes #940.

May be a little heavy-handed since it just marks all layers dirty I think. Only marking the layer dirty where the preview was removed from doesn't work though, so this is my solution for now.

### DIFF
--- a/src/libclient/core/layerstack.cpp
+++ b/src/libclient/core/layerstack.cpp
@@ -742,7 +742,9 @@ void EditableLayerStack::reset()
 void EditableLayerStack::removePreviews()
 {
 	for(Layer *l : d->m_layers) {
-		EditableLayer(l, d, contextId).removePreviews();
+		EditableLayer el(l, d, contextId);
+		el.removePreviews();
+		el.markOpaqueDirty();
 	}
 }
 


callaa:
I think I see the bug. When `setHidden` is called on a layer, it internally calls the markOpaqueDirty function, but it's called on the temporary sublayer and not the actual parent layer. I'll have to give this some additional testing, because I'm not sure if calling markOpaqueDirty on the parent layer after hiding the sublayer works in all cases.

callaa:
Okay, I think I figured out the bug. When an undo is performed, the layer stack is reverted to a previous copy, which does not include the temporary layers. Thus, the when removePreviews is called, the temporary sublayer layer that is used to mask away the source of the move no longer exists, thus markOpaqueDirty is never called.

Your solution is a simple way to fix it, and probably the best cost/effort tradeoff.

PR #945, opened 2020-12-27
Allow different input settings per brush

### LINKED ISSUE #915: Feature Request: Smoothing per brush?
Possible to maybe have the smoothing be applied per brush if possible? Sometimes smoothing works great for drawing, but then for details or writing, it tends to interfere.

askmeaboutlo0m:
The input tab now allows specifying multiple different smoothing and
pressure mapping configurations. Every brush slot lets you pick which of
these configurations you want to use for it.

This lets you, for example, create a "sketch" configuration with little
smoothing assigned to brush 1 and a "line" configuration with lots of
smoothing for brush 2.

Resolves #915.

Showing it off:

https://user-images.githubusercontent.com/13625824/103174749-d4db3f80-4864-11eb-87d5-2b4ffa7f9253.mp4

### DIFF
--- a/src/desktop/CMakeLists.txt
+++ b/src/desktop/CMakeLists.txt
@@ -12,6 +12,7 @@ set (
 	main.cpp
 	mainwindow.cpp
 	notifications.cpp
+	inputpresetmodel.cpp
 	toolwidgets/toolsettings.cpp
 	toolwidgets/brushsettings.cpp
 	toolwidgets/colorpickersettings.cpp
--- a/src/desktop/docks/inputsettingsdock.cpp
+++ b/src/desktop/docks/inputsettingsdock.cpp
@@ -22,16 +22,16 @@
 
 #include "ui_inputcfg.h"
 
-#include <QSettings>
+#include <QDebug>
+#include <QInputDialog>
+#include <QUuid>
 
 namespace docks {
 
-namespace {
-	int center(const QSlider *slider) { return (slider->maximum() - slider->minimum()) / 2 + slider->minimum(); }
-}
-
-InputSettings::InputSettings(QWidget *parent) :
-	QDockWidget(tr("Input"), parent)
+InputSettings::InputSettings(input::PresetModel *presetModel, QWidget *parent) :
+	QDockWidget(tr("Input"), parent),
+	m_presetModel(presetModel),
+	m_updateInProgress(false)
 {
 	m_ui = new Ui_InputSettings;
 	QWidget *w = new QWidget(this);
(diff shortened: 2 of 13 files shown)


callaa:
Thank you, this looks like a great feature. Since it's a bigger PR, I'll need some time to go through it.
I'm sorry I haven't reacted sooner to any of these. What was supposed to be just a short break turned into a long hiatus, but I'm trying to get back into the routine now.

callaa:
Thank you, I've looked through the pull requests and the feature is certainly worth including.
You can put your name in the copyright section of the files you've created.

There are a few improvements/changes I'd like to make, if that's OK:

 - The input dock is getting rather crowded now. Vertical space is always at a premium, so this is a concern.
   However, since there is now an input preset selector in brush settings, there isn't a need to use it very often, so it
   could be hidden by default. (Or maybe even replaced by a dialog box?)
   We could also have more default presets: stylus, distance and velocity, perhaps.
 - The dropdown in the input dock could be a combobox like in the palette box, so there isn't a need to have a separate "rename" action
 - From a quick glance, the PresetCreator abstract class isn't really necessary. Rather than using factory classes, we could just use Preset as a value type and add a static fromSettings method.

notfood:
Rather than a combo box, it could use a widget like brush selection with 1 to x numbers. Usually you won't have that many input curves and a single click is better than two.

Or take the Krita approach and make the input curve per dab setting which would open up more brush styles that aren't possible right now.

askmeaboutlo0m:
I kinda like the combo box better because I can name the presets, rather than having those cold, impersonal numbers that I have to go back and check what they mean. So I wouldn't voluntarily go back and put effort into re-jigging it into a system that has arbitrary limits instead.

I don't know what you mean with the second paragraph, but I also don't understand Drawpile's brush dialog with all those scary, unlabeled brush circles in them that don't seem to do anything.

callaa:
I rebased this PR to master (and also moved the inputpresetmodel.{cpp,h} files to the src/libclient/canvas/ directory. Still perhaps not the ideal place for it, but seemed like the best choice for now.) I pushed it as the `pr945` branch.

I'd like to implement the changes I proposed, but I'd like your opinion on one thing first. Presently, there are separate curves for each pressure mode: stylus, distance and velocity. I wonder if this should be simplified now that we have proper presets, so each preset would only have one curve. Looking at this from the "principle of least surprise" perspective, it seems more intuitive to me that changing the input type wouldn't also switch the curve.

callaa:
Almost forgot one other: the input dock feels rather separate from the brush settings dock and it might not be obvious to new users where the presets are edited. I wonder if it would be a good idea to turn the input dock into a popup dialog that is opened by clicking a "edit" button that could be added next to the preset choice box. This would free up some vertical space and allow the editor box to be bigger.

askmeaboutlo0m:
Moving the input to a popup seems fine, the input tab is notoriously hard to find, I've had several people ask me where to change their pressure stuff. But then maybe "input" or "preset" should be changed to something more obvious. Dunno what the right term is though. Maybe stabilizer? I mean, that's not really the correct term, but it's what people seem to be looking for.

It was a combo box initially actually, but I really, really hated the way editing worked (it didn't), so I changed it to be the separate dialog instead of fighting with QLineInput event filtering and whatnot. Maybe you can make it work properly though, I don't really know Qt at all.

The PresetCreator class is just a result of refactoring duplicated logic, it reduced code size by a bunch. If you have an idea how to make it even less code, sure!

Having a single curve instead of three seems much more sensible, yeah. No reason why they should be split anymore. Does the same go for the vertical slider next to it? I don't actually know what that does since I never use those other dynamics.

callaa:
I just pushed a couple new commits to the pr945 branch. You can give it a test if you like.
I did some general refactoring and got the combobox editing working. I understand what you mean though, working with Qt's model classes is tough. Even though I've made quite a few editable models in Drawpile, I still fought for like 30 minutes with this one because I had made a tiny little copy&paste error.

In any case, I've made the following behavioral changes:

 * There is now a single curve per preset, so changing the stylus/distance/velocity mode no longer changes the curve
 * I removed the "rename" action, since renaming can now be done directly in the combobox. But I added a "Duplicate" action, so it matches the palette box hamburger menu.
 * I removed support for legacy input curves. (I might add back a legacy->preset migration before merging this branch, though.)

Also one important change: There is now a PresetModel singleton that is used by all windows. This is important, because otherwise it's possible for changes to be lost if more than one window is open.

Things left to do:
 * Move the input dock to a popup window opened via the brush editor. (I already noticed it's confusing to edit a curve and have nothing happen because another curve was selected in the brush.)
 * Code for migrating legacy input settings into presets
 * Maybe an import/export feature for presets, like there is for palettes?

callaa:
This change is now merged.

notfood:
Heads up. This caused https://github.com/drawpile/Drawpile/issues/954 issue

PR #947, opened 2021-01-03
Add perspective transform (distort/free transform/free deform)

askmeaboutlo0m:
A feature that goes by many names in different programs. Lets you do a perspective transform on a selection, something I use all the time.

Also makes some usability improvements to the selection tool panel because I was going slightly mad testing this out. It now disables all the stuff in it when no selection is active and provides a set of radio buttons to switch between the different transformation modes.

I'd like to make more improvements to this to basically make it work more like it does in Krita (which e.g. doesn't have the issue of accidentally clicking outside the selection), but that's unrelated to the direct functionality here, so I think it could be merged already and hopefully be as useful to others.

Showing it off:

https://user-images.githubusercontent.com/13625824/103480108-94cf0c00-4dd2-11eb-9f28-e82264635b0b.mp4

### DIFF
--- a/src/desktop/scene/arrows_data.h
+++ b/src/desktop/scene/arrows_data.h
@@ -41,6 +41,30 @@ static const QPointF
 ,verticalSkew[] {
 	{2,5}, {8,0}, {8,5}, {2,10}
 }
+,distortTopLeft[] {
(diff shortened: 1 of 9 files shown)


callaa:
Merged manually.

PR #966, opened 2021-06-28
Add HSV Blend Modes: Hue, Saturation, Luminosity and Color

askmeaboutlo0m:
Whoops, I accidentally closed #921 because I kept it in the master branch. Looks like I can't reopen it, so this is just a repeat of the information there.

I recently learned of a painting technique where you paint everything in grayscale first and then add the color later. I quite like that method, since it lets me split up the cognitive load of figuring out the shapes and the colors separately. However, Drawpile doesn't support a "Color" blend mode, which is kinda necessary to work with this technique.

This pull request implements that Color blend mode together with the related HSL modes Hue, Saturation and Luminosity. Those are all supported by the SVG spec as well, so I think it's reasonable to add them all.

This tragedy comes in three parts:

1. (No longer relevant, already merged.) Making this build on my system. I'm on Arch Linux, so I think my library versions are newer than they are for someone on a reasonable distribution. I've had to #include <QPainterPath> in some places and add compatibility for the MHD_Result enum introduced in libmicrohttpd 0.9.71.

2. Adding the actual blend modes. Implements the HSL blending and changes around some of the rasterop functions to avoid duplicating the pixel composite functions. It's still template functions generated at compile time though, so no added dynamicity that could slow this hot code down.

3. Make the brush preview more useful for these modes. Normally, the rainbow dabs in the preview only vary in their hue, but for these HSL modes, they vary their saturation and value as well.

I've provided details in the individual commit messages as well.

This will have to wait until 2.2 because of protocol changes.


askmeaboutlo0m:
Closing in favor of #1038.

Issue #681, opened 2019-01-30
feature request: hcl color picker


gebgebgeb:
it would be great to have the option of using hue chroma luminance sliders for picking colors!

PR #967, opened 2021-06-28
Add colorwheel shape, angle and space options

askmeaboutlo0m:
A new tab in preferences allows you to set these and wires them up to
the matching options in the ColorWheel class. If no settings exist yet,
it'll default to a rectangular, fixed colorwheel with a square shape, as
before.

From my testing, the Lch color space acts in a really confusing manner.
I guess this would make sense to people who know how to work it, but I
dunno if it's actually a use case for Drawpile, so maybe it could be
removed to avoid bug reports by users that are as perplexed by the
behavior as I am.

Attached is a video of it in action.

https://user-images.githubusercontent.com/13625824/123676494-4ca21b00-d844-11eb-825e-f35697bc9014.mp4

### DIFF
--- a/src/desktop/dialogs/settingsdialog.cpp
+++ b/src/desktop/dialogs/settingsdialog.cpp
@@ -53,6 +53,8 @@
 
 #include <QDebug>
 
+using color_widgets::ColorWheel;
+
 class KeySequenceEditFactory : public QItemEditorCreatorBase
 {
 public:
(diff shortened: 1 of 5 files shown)


askmeaboutlo0m:
Looks like this would fix #681, which asks for the Lch option.

askmeaboutlo0m:
Fixed an issue where the color wheel would be super tiny on Windows, instead of expanding to fill the preview panel. Now it's just set to a reasonable minimum size.

callaa:
Very nice, thank you!

PR #968, opened 2021-07-01
Improve input presets dialog

askmeaboutlo0m:
This amends changes made by @callaa to move the input presets from their own tool thingy into a dialog. Adds a few things to make it look and behave reasonably.

* Sensible padding around the edges and between its elements, in line with other dialogs.

* A minimum size for the bar on the left so the handle doesn't get clipped on the left.

* Expanding vertical layout policy for the curve. Otherwise hiding the bar on the left will cause it to collapse.

* A close button at the bottom. It would be more in line with other preferences to have this be a set of OK and Cancel buttons instead, but since you probably want to be trying out your input settings while you're adjusting them, this is probably fine.

* A tooltip on the edit presets button.

* Hiding and showing the slider bar on the left depending on which input mode is chosen. This slider does nothing for stylus input, so it's hidden for that one, the other two have it shown

* Said slider also gets proper tooltips now, talking about distance and velocity depending on which is selected. Before it always talked about velocity.

This is what it looks like:

https://user-images.githubusercontent.com/13625824/124179265-dfe08800-dab2-11eb-8b6e-bb470c11c6a2.mp4

### DIFF
--- a/src/desktop/dialogs/inputsettings.cpp
+++ b/src/desktop/dialogs/inputsettings.cpp
@@ -126,6 +126,7 @@ void InputSettings::applyPresetToUi(const input::Preset &preset)
 		m_ui->pressuresrc->setCurrentIndex(preset.curve.mode);
 		m_ui->curve->setCurve(preset.curve.curve);
 		m_ui->curveParam->setValue(int(preset.curve.param * 100.0));
+		updateModeUi(preset.curve.mode);
 		m_updateInProgress = false;
 	}
 }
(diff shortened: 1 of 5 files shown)


askmeaboutlo0m:
Amendments: the delete button now gets enabled properly when new presets are added and the input preset dialog is kept in sync with the current brush so that you don't end up confused why nothing changes as you're trying out your settings until you realize the preset you're editing isn't the currently selected one.

Issue #836, opened 2019-11-22
Brush Puck replaced by wrong cursor when leaving canvas area


redrodent:
iMac, up to date. Drawpile, up to date. When leaving the canvas area of the screen and going onto one of the windows with tools or brush options, etc. The painting puck gets replaced by either a regular mouse cursor, or the tools that let you adjust the size of your tool windows. The only workaround I've found is tabbing out, which is inconvenient and only fixes it for as long as I stay on the canvas.

Screenshot attached with a drawing of the issue since I can't screen cap my cursor....

Basically it looks like Drawpile forgets that I want to be using the drawing/painting puck when I go back onto the canvas from my toolbars/windows.

![Screen Shot 2019-11-21 at 8 50 24 PM](https://user-images.githubusercontent.com/10472724/69398784-8f7f3280-0ca0-11ea-879b-af1791886830.png)

Tabbing out and back into Drawpile fixes it, but not permanently.

redrodent:
Taking my cursor far enough away from my Wacom Cintiq display also resets the cursor.

Wade821:
Thanks for alerting us of the issue. When providing information about your operating system and drawpile version however, please provide specific numbers. We don't know if latest means the latest your computer can get because it's old, or if it's the newest release from Apple. 

Same with Drawpile. Some people update a month ago and don't actually check for updates. Oof. Giving numbers assures that we're on the same page.

dustiibunnii:
Oh! This happens on my MacBook Pro as well. Currently I am using a Catalina 10.15.5, with drawpile 2.1.17, and whenever I go off screen, and then back in, the circle will be overlapped by the regular curser, or if I was typing in chat, by the ‘I’ symbol. It doesn’t appear to interfere with the strokes, it’s just the curser is changed.

userwer:
I am experiencing the same issue. Whenever my cursor leaves the canvas area, it will stay in the cursor shape upon returning, and most of the time the brush/eraser outlines do not reappear. Sometimes (I am not sure exactly when) both the cursor and the outlines are there together. I am able to temporarily resolve the problem by clicking on a drop-down menu (like the edit menu, or the zoom menu where it lists percentages), and clicking directly into the canvas with my pen. The brush and eraser outlines will be back to normal until my cursor leaves the canvas again.

[...]

PR #969, opened 2021-07-01
Implement canvas viewport entry hack

askmeaboutlo0m:
On my Linux system, when entering the canvas viewport via tablet, the cursor doesn't update properly. This implements a workaround that allows Qt to generate synthetic mouse events for tablet move events, which function properly. The canvas view filters out these synthetic events when the hack is enabled, which means that there should be no change in behavior.

For reference: this is Arch Linux with whatever current Qt5 it comes with, the issue has been there since forever I think. This might be worth raising as an issue with Qt itself, but that'd need a more minimal example to show off the issue in isolation.

This video shows the issue in action. When I move the cursor into the canvas viewport with my tablet, it remains a regular arrow cursor and doesn't have the rectangle tool symbol next to it. After enabling the hack, the cursor behaves properly.

https://user-images.githubusercontent.com/13625824/124194353-b7fc1f00-dac8-11eb-9e22-e32c63922ec4.mp4

### DIFF
--- a/src/desktop/dialogs/settingsdialog.cpp
+++ b/src/desktop/dialogs/settingsdialog.cpp
@@ -124,11 +124,6 @@ SettingsDialog::SettingsDialog(QWidget *parent)
 	m_ui->relativePenModeHack->hide();
 #endif
 
-	// Same for Linux-specific hacks.
-#if !defined(Q_OS_LINUX)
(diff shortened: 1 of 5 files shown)


askmeaboutlo0m:
Okay, according to #836, this happens on other operating systems too. So I removed the Linux-only restrictions.

PR #981, opened 2021-07-30
brushPalette: prevent deleting Default folder, preventing crashses when no folders exist and using certain actions

taiyu-len:
When deleting all folders in brush presets, the addBrush, and deleteFolder menu actions will cause drawpile to crash
This is adds a few checks, along with qWarning when using these actions with no folder.

Just saw the comment `// The first folder is always the built-in "Default" folder` so i assume that that folder should not be deleted and will add some checks, and such for that case

---
- Old

Before merging I want to ask, what would the best choice be to handle this situation

1. disable invalid menu options
2. hide invalid menu options (only addFolder  would show in context menu)
3. prevent last folder from being deleted (Alternatively, prevent Default folder from being deleted)

I have 1 ready to push, and 2 requires only a small edit.

Considering how you cannot rename default, and if you delete it and create a new one, (new folder) you cant rename that,
it also matches behavior from the input presets, so 3 seems best i guess

### DIFF
--- a/src/desktop/docks/brushpalettedock.cpp
+++ b/src/desktop/docks/brushpalettedock.cpp
@@ -100,6 +100,12 @@ BrushPalette::BrushPalette(QWidget *parent)
 	auto *addFolderAction = hamburgerMenu->addAction("New Folder");
 	auto *deleteFolderAction = hamburgerMenu->addAction("Delete Folder");
 
+	// Default folder cannot be deleted
+	deleteFolderAction->setEnabled(false);
+	connect(d->ui.folder, QOverload<int>::of(&QComboBox::currentIndexChanged), deleteFolderAction, [=](int index) {
+		deleteFolderAction->setEnabled(index != 0);
+	});
+
 	d->ui.menuButton->setMenu(hamburgerMenu);
 
 	d->ui.brushPaletteView->setContextMenuPolicy(Qt::CustomContextMenu);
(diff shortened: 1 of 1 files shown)


taiyu-len:
Ok, changed it to prevent default folder from being deleted, disabling the menu action when default is selected, and a check to prevent it from being deleted via deletefolder

callaa:
Thanks! I think your proposed solution number three would be better, since it doesn't matter which brush folder exists, as long as at least one does. This should be an easy change: simply check the number of items in the folder list rather than the current index.

taiyu-len:
Sure, however drawpile makes a few assumptions that Default is the first folder in a few places so that might need to be changed
https://github.com/drawpile/Drawpile/blob/ee4f123e21e4bf04e2b4c6ae0a950e38cb91d93b/src/libclient/brushes/brushpresetmodel.cpp#L223-L231
https://github.com/drawpile/Drawpile/blob/ee4f123e21e4bf04e2b4c6ae0a950e38cb91d93b/src/libclient/brushes/brushpresetmodel.cpp#L556-L557
https://github.com/drawpile/Drawpile/blob/ee4f123e21e4bf04e2b4c6ae0a950e38cb91d93b/src/desktop/docks/brushpalettedock.cpp#L62-L63

just need to change these few areas to allow arbitrary filename for the first folder then

callaa:
Oh, right! In that case, the PR is good as it is.

PR #985, opened 2021-08-05
Update QtColorWidgets

### LINKED ISSUE #984: HSL and LCH color wheels have their values locked
Unlike the HSV color wheel, HSL and LCH values lock whenever you set it to white or black.

HSV:
HSV

HSL:
HSL

notfood:
- Split in two commits so it's clearer.
- Moved the headers files to bundled/QtColorWidgets/include to match the structure of upstream, keeping them in the same folder would have required to change a lot of files.
- Config settings is now 3 values: `shape`, `rotate` and `space`
- Removed redundant `SettingsDialog::restoreColorWheelSettings(int flags)` and put the logic in the value setters.
- Removed redundant enums `ColorWheelShape`, `ColorWheelAngle` and `ColorWheelSpace` as they're the same as the ones in QtColorWidgets, with the minor exception of `ColorWheelShape`, which has the order inverted.
- Fix the names of colorwheelShapeBox to reflect the enum change.

Fixes #984


callaa:
Thanks. Since this pulls in the whole package, I'll look into if its worth replacing Drawpile's own palette models with the ones provided by QtColorWidgets.

PR #998, opened 2021-09-13
Rust: Add Screen Blend Mode

askmeaboutlo0m:
* Change the Divide mode SVG name to -dp-divide

Rather than screen, because Screen and divide are totally different blend modes (one is useful being the opposite of Multiply, the other is Divide.)

* Use "composite-op" attribute in ora writer

For layer modes, instead of "blendmode". Otherwise nobody will understand our ora files, not even our own ora reader!

* Add a Screen blend mode

The opposite of Multiply: it lightens stuff. Saving as ORA and loading that in Krita gives the same result, so looks like it's implemented correctly.

Supersedes #939.


callaa:
Great! Do you suppose the Divide mode is useful at all? Now that we have the Screen mode, should it be removed since it doesn't even have a SVG name?

PR #1005, opened 2021-11-07
Update mainwindow.cpp

chainofflowers:
FIX document name in window's title

Some documents may contain dots in their name, which are actually part of the name. Only the last dot should be considered the separator from the "filename extension" (Drawpile does not open files with multiple dots in the filename extension, like for instance *.tar.gz).


chainofflowers:
Hi @callaa ! Anything against merging this? :-)

callaa:
Thanks! Sorry I didn't respond sooner.

PR #1037, opened 2022-08-13
MyPaint Brush Support

### LINKED ISSUE #179: Feature: real stabilizer for classic brushes

### LINKED ISSUE #498: minimum/maximum brush size?
A feature I find myself missing from other programs is the ability to set a "minimum size" for a brush, ala PTS, so that I even with the lightest stroke, I wind up with say, a 25% pressure line. Is this something that would work well in drawpile, or would this end up making the UI too cluttered? Maybe something gated behind a checkbox or a menu?

### LINKED ISSUE #617: Custom Brushes Feature
hi there! So I am here playing about with Drawpile and figuring out new things everything! 

However, there are times where I want to start painting. Of course I can start using soft brushes to help me paint, but I kind of thought about if there is a way to make custom textured brushes? I feel like that would be an amazing feature if its added! I would love to collaborate with people online with custom brushes.

### LINKED ISSUE #662: [Feature request] Blur tool, whisper for chat, friend adding
The new changes to the watercolor tool are amazing, but I was wondering if there could be a blender blur tool that was somewhat like the ones in Krita?

It would also be a cool feature to be able to add friends and whisper in chat.

### LINKED ISSUE #800: Feature request: Brush Preset naming and preset stroke preview.
drawpiles brush preset browser in its current state is somewhat useless when dealing with lots of saved brushes. since theres no way to tell the difference between certain brush configs. having the option to name presets will fix this issue. having a visual brush stroke preview in the browser would be a nice addition aswell.
Captuggre

### LINKED ISSUE #933: Add more stabilizer options or more levels of smoothing
The current maximum level of Smoothing from Drawpile is just not enough to stabilize my strokes, it makes doing a proper cartoony lineart really hard to do!

Can you add more levels of stabilization to the smoothing? Or override options, or just allow us to pick a custom smoothing level, I really want the stabilizer to have additional levels, at least something closer to Paint Tool Sai S-2 stabilization or the value 30 on Clip Studio's stabilizer setting.

I actually tried to use 3rd party software like Lazy Nezumi and it doesn't work... I think it's just better to implement this feature directly on Drawpile or just updating the Smoothing slider with even more smoothing

### LINKED ISSUE #944: Bug: Smudging Tool
Hello!  I have a few features to request, or suggest!

- The brush system here is a bit difficult to grasp, however after a bit most get it. Though, I was wondering if we  could rename presets or something like that, or even, have the paint system work like firealpaca or paint tool sai, photoshop--  more like, having a list of brushes and their names, which changes their presets, and even having some kind of way to create our own textured brushes and add them, group via folders... You know how firealpaca can upload bitmap files and alter settings in a brush zone? Or even upload code files for those patterned brushes.

- Folder support for layers. I am not sure how feasible my thoughts are, but I thought I could put it out. Perhaps there could be a way to drag and drop layers into folders, collapse them, and on top of being able to control each individual opacity, control the grouped opacity.

I do realize that drawpile is orientated for the mass user, and the drawing on one canvas thing, however I thought these would  allow users to get more comfortable with being able to use brushes that are more suited to be added/removed, named, customized, etc... As well as, the  blur/smudging tool seems to  be buggy? For me, it creates this weird rainbow-edged glow smudge when it's a really large brush size.

Also, I did read about  someone  doing  something on  Rust, and revamping the core Drawpile thingos- that's pretty awesome.
Thank you for considering this if you do!

askmeaboutlo0m:
libmypaint comes bundled, so the build is still the same as it was.

Fixes #179 and fixes #933 - MyPaint brushes have a stabilizer.

Fixes #498 - sorta anyway, MyPaint brushes can do this anyway.

Fixes #617 - oh, hi Ben, here's cool brushes.

Fixes #662 - there's smudge and blur brushes now. They have trouble with opacity, but so does everything with Drawpile due to its 8 bitness, so that's a different issue.

Fixes #800 - presets docker has been reworked to support naming, tagging and searching.

Fixes #944 - this is like three issues in one, but they all get fixed by Drawpile 2.2 and this I think.

## Move opacity from masks into raster operation

MyPaint brushes can draw the same dab multiple times, each with a different opacity scale. To avoid regenerating the mask just to change its baked-in opacity, the opacity now instead gets multiplied into the blending operations. That's also how MyPaint does it. I guess it would also make it easier to cache brush dabs, since they're now less varied.

## Add a delta time parameter to brush strokes

MyPaint requires this for the sake of smoothing and stuff. Classic brushes just ignore the parameter.

## Implement converting Rustpile color to HSV

Required for MyPaint brushes, since they only save their color in that format and we want to reconstitute that somehow. Would also be useful for HSV blend modes.

## Implement NormalAndEraser blend mode

Used by MyPaint brushes for smudging and such.

## Let cbindgen include ostream and use constexpr

I dunno, I regenerated the rustpile bindings and it gave me this. Probably just what newer versions of cbindgen do.

## Add a bundled libmypaint and Rust bindings for it

Only pulling in the bits of libmypaint that we actually need and patching out its useless dependency on json-c. That means no further setup is necessary, it just builds out of the box.

## Implement MyPaint brushes

Drawing brush dabs, previews, brush settings docker, dabs generation and the new dab messages as well as the handling thereof.

Reworks a lot of places where ClassicBrush was used into an ActiveBrush that can toggle between that and MyPaint brushes.

## Smooth movements of usermarkers by default

MyPaint brushes can cause dabs to appear in wildly different places, causing the user marker to jerk around like crazy. This commit makes the markers interpolate their position instead, meaning that they're less accurate, but also way less annoying.

Laser pointer markers still work as before, since their accuracy is more important and they don't have the crazy dabs issue in the first place.

This speeds up the animation timer tenfold, since otherwise the movement is too inaccurate. It's also definitely a hack, but it seems okay to me, the position of the user markers is still close enough. Maybe it could be restricted to only to the interpolation when a MyPaint brush is used, but that might be too much effort for something that's barely an issue.

What I definitely would *not* do is introduce new messages to report the cursor position, the markers aren't worth that.

## MyPaint support and rework of brush presets docker

In addition to supporting MyPaint brushes, it now behaves more like Krita's presets docker in that it uses tags instead of folders. A single brush can be put into multiple tags. Brushes can also be named, given a description and a custom thumbnail now. Sort order is now dependent on name, which is what MyPaint and Krita do, rather than having the user push them around. Searching of brushes is also now supported, since they got names to search them by. And of course you can import MyPaint presets.

The internals are reworked to just use SQLite instead of dealing with bespoke files and messing around with that timer. Old presets are converted to the new format on the first run through.

## Add MyPaint brush presets database

In the form of initialbrushpresets.db, which is shipped with Drawpile. On first launch, it's copied to a writable location and used further from there.

## Add myself to the AUTHORS file

Forgot about doing that earlier, but now I actually exist.

## Update old callaa/Drawpile wiki link in README

To point at drawpile/Drawpile instead.

## Add Python venv directories to .gitignore

Since that's useful for dealing with the protocol generation and not having to install Python packages into the system environment.

## Add clangd compile_commands.json to .gitignore

When using -DCMAKE_EXPORT_COMPILE_COMMANDS=ON, you get autocompletion support for clangd, but have to put the file (or a symlink to it inside the build directory) into the root directory for it to work properly, so on the ignore list it goes.


askmeaboutlo0m:
The force pushed additions are:

* A rebase on dev-2.2.
* Build libmypaint through Cargo.
* Fix a smudge bucket bug in libmypaint.
* Separate strokes in MyPaint properly.
* Fix the Normal and Erase blend mode to use unpremultiplied colors.
* Allow picking icon sizes in the new brush preset docker.

PR #1043, opened 2022-08-31
MyPaint Brush Improvements

askmeaboutlo0m:
Supports colorize and posterize for MyPaint brushes. This breaks protocol, so we should probably increment the version? I'll leave that up to you.

Removes the Pigment slider, because that's spectral painting stuff, which we don't support and now force off in MyPaintBrushState. It's some weird realistic-ish color mixing thing with matrices that's really brutal on performance and cranked up really high by all MyPaint brushes by default. I don't think it's worth supporting and making everything slow for it.

Now drops dabs of infinitesimal size, opacity or hardness, like MyPaint does it.

Makes the freehand tool actually use the MyPaint brush size when it's selected, before it only applied it sometimes.

And finally, supports pen tilt and barrel rotation for MyPaint brushes, for people with tablets that report that kinda thing.

Edit: and also add a tooltip to the brush preview that says how many dabs it drew, to help users get _some_ kinda idea of how "heavy" a brush is. Some of them draw orders of magnitude more dabs than a classic brush, causing resets really fast.

Edit²: create the parent directory of the brush presets database if it doesn't exist yet. This happens if it's the first time a user has ever run Drawpile.

### DIFF
--- a/src/desktop/scene/canvasview.cpp
+++ b/src/desktop/scene/canvasview.cpp
@@ -416,12 +416,12 @@ void CanvasView::leaveEvent(QEvent *event)
 	updateOutline();
 }
 
-canvas::Point CanvasView::mapToScene(const QPoint &point, qreal pressure) const
+canvas::Point CanvasView::mapToScene(const QPoint &point, qreal pressure, qreal xtilt, qreal ytilt) const
 {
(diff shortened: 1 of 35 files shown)


askmeaboutlo0m:
Closing because of Dancepile taking over.

PR #1050, opened 2022-10-03
Add a forceNsfm option for the server

askmeaboutlo0m:
Some servers mandate the use of NSFM on all sessions. This adds a `forceNsfm` config option that enforces this mandate automatically, it makes it so that all sessions are set to NSFM and can't be unset.

This is split into two commits for easier cherry-picking. The server part applies cleanly to master as well, so if a server operator wants to build a server with it, they can do so by running the following:

```bash
# Add my Drawpile repository.
git remote add loom https://github.com/askmeaboutlo0m/Drawpile.git
# Grab the branches from it.
git fetch loom
# Apply the commit to your current state, presumably the official master branch.
git cherry-pick 9e9488a80650984eb784f567c6882b94bd37cd08
```


askmeaboutlo0m:
Closing because of Dancepile taking over.

PR #1057, opened 2023-01-05
Fix macOS build

csnover:
Since the appbundle target copies from files that are generated by the dpclient dependency it must be built first.


csnover:
This is superseded by #1060.

PR #1058, opened 2023-01-05
Fix crashes when sampling the edge of a layer

csnover:


askmeaboutlo0m:
Thanks for your work. The `dev-2.2` branch has been subsumed by the `dancepile-2.2` branch, which replaces the Rust paint engine with [Drawdance](https://github.com/drawpile/Drawdance/tree/dev-2.2). I'll still see about merging these pull requests, but as it stands I don't think there will be further developments on it.

PR #1060, opened 2023-01-13
Rewrite CMake files to generate working projects and enable GitHub Actions for CI

csnover:
This patch series was created after attempting just to get a working Drawpile client build and struggling a lot because the build system seemed to have ossified a bit and didn’t work with anything other than Unix Makefiles. No one usually likes working on build systems so I figured I would just go ahead and try to fix it.

Because the CI has not been working since Travis CI discontinued their open source support, there were various compiler warnings, linting warnings, and broken tests that needed to be fixed (broken out by commit). There was also some unnecessary unsafe code on the Rust side that I also just went ahead and replaced because I noticed it was there and it was trivial to fix.

This patch set also includes preliminary (untested—it compiles now and that’s all I’ve checked so far) support for Qt6, though there are still some more deprecation warnings to address. It establishes a minimum Qt version of 5.11 for the server code (because it was trivial to support and Debian still uses Qt 5.11 in supported LTS releases), and 5.12 for the client code.

The most relevant changes are described in the commit message for the CMake commit, so please read that to understand the changes. I tried to avoid making any change that would be particularly disruptive and hopefully the only things that are breaking changes are (1) the requirement to use an out-of-source build, and (2) the requirement to specify explicitly the build type when running CMake for a single-target generator.

I have tested this new CMake build generator with VS2022 on Windows and Xcode on Mac, as well as the generators that are used by the CI, and it appears to correctly generate working builds on all of these platforms.

In future PRs I plan to make these changes to CI:

1. Use something like vcpkg to get rid of platform-specific dependency management in the workflow (especially since Ubuntu apt repository is flaky and the Windows CI is relying on outdated prebuilt binaries from Qt Company). Currently this is not the case because it was faster to get CI up and running with third-party binaries than to also try to make it work whilst waiting for long uncached Qt builds to fail.
2. Compile with `-Werror` once the remaining Qt6-specific warnings are fixed.
3. Integrate installer generation and publish release binaries automatically when tags are pushed so no other special infrastructure is required to generate release versions or test builds.

Additions like clang-lint would also probably be a good idea, but I don’t have an immediate plan to add that.

I will be submitting a couple more patches that will be based on top of this one for your consideration soon. Thanks!


csnover:
This patch series has been updated. The big additional changes since the original PR are:

CI builds run against latest releases of Qt5 and Qt6. Qt6 now works without deprecation warnings. A compatibility header has been added to libshared to facilitate future Qt updates in a clean and clear way.

The filesystem has been cleaned up so that everything is in a place that is most appropriate, instead of having source packages that reach into the root and vice-versa. Graphics assets have been deduplicated and missing ones added. Values that were hard-coded or that existed in multiple places have been unified; this means that things like man pages, systemd units, file associations, etc. are now fixed so that they will not so easily become out-of-sync. (WebP support was also added in this change since this part of the code was being touched anyway.) Includes are normalised so all in-source include paths are relative to `src` and current directory is no longer included in search path. This means this patch set will bitrot relatively quickly if other changes are committed before it is committed.

Complete WIP and release builds can now be generated by CI for Windows, macOS, and Linux. This uses source builds for all of the libraries that are linked directly to Drawpile. The generated builds are of type RelWithDebInfo. Because the external network access on CI can be flaky, it caches all dependencies that would normally cause a network request. There are a few notable caveats:

* Building on CI is a little slow mostly because LTO is enabled all the time; this should probably be updated so that LTO is only enabled for release builds, but this is outside the scope of this PR for now, as it is already too big and I want to be done with it.
* The AppImages are built by the latest Ubuntu and do not force old glibc symbol versions so are not good universal AppImages. I looked into using a different tool to generate these images that would include a compatible glibc, but the ones I saw did not seem to have all the needed features to handle multiple executables and correct auto-resolution of library paths. Something like compiling with one of the pre-built musl compilers might be a better solution here since then it will use the always-stable musl ABI instead, but I didn’t feel like continuing to investigate. The AppImage generated in CI was tested successfully on a fresh Debian testing VM with a non-KDE DE, so it is at least somewhat portable.
* Because CPack is not good and does not seem to install components in a correct order (so runs macdeployqt/windeployqt/etc. before all targets are copied into the installation directory), Windows installer releases are currently built as a monolithic release. This is not a large burden as these are only a few megabytes anyway; it may not be worth it to try to componentise the installation at all.
* The Windows installer is now an MSI installer using WiX Toolset via CPack. This reuses the existing GUID of the earlier Inno Setup installer; this may not be desired, in which case it can be replaced with a new GUID. It has been tested successfully on fresh Microsoft Windows 11 WinDev2301Eval VM.
* The Windows ZIP will usually try and fail to write to the user’s roaming storage; I assume this is some Windows security feature but I do not normally work with Windows so I don’t know what is the deal with that. It does work as a portable build using the already existing command-line flags for that.

Release management is now almost fully automated using the `pkg/release.sh` script. This script runs all of the necessary updates to appdata.xml, ChangeLog, etc., appropriately creates commits and tags, and pushes to GitHub, where the GitHub Actions workflow will then publish automatically using data from the ChangeLog for the tagged version. The only remaining manual actions for releases are web-site–related and Docker-related. All of the old packaging stuff has been removed as it is no longer needed.

Because the auto-update system is so closely tied to the build system (using appdata.xml, taking version number from config.h, etc.), it has been updated. It is now compatible with semver and much harder to break with incompatible data changes. Correct platform triplets are used now when checking for binary updates and Linux binary updates are (ostensibly; this is untested) supported. Users can now also opt in to receiving beta updates in the Network settings, auto-update settings are visible in Network settings, and are asked if they want auto-update enabled the first time they ever run Drawpile.

Due to the use of invalid platform triplets in earlier releases that cause AppImage generation to fail, appdata.xml in the source tree will not work with Drawpile 2.1 auto-update any more. To deal with this issue, `update-appdata-releases.py` has some new command-line options that allow it to generate Drawpile 2.1-compatible release entries that can be uploaded to the web site. Using appdata.xml is probably not a good idea anyway since (1) it is a relatively inefficient way to retrieve updates as it includes all releases; (2) the format is controlled by a third-party; (3) the format is not designed explicitly for cross-platform application auto-updates like this. However, no changes have been made to replace it.

Auto-update will now be turned off if the server responds to a request with a 410 status code, or if the client has been unable to contact the update server for more than a month. (I learned the hard way once that without this kind of feature, old clients will keep hammering the server for updates until the end of time, with no way to get it to stop.)

`-Werror` is enabled now on all compilers and the necessary code fixes to enable this have been made.

Looking forward to any feedback. I will probably start generating new releases on my fork for anyone who is interested in my other fixes, since there doesn’t seem to be any activity or feedback occuring upstream here for half of a year now. I would be happy to temporarily act as maintainer on this upstream fork if this is desired in order to get all these fixes and enhancements landed and a new release pushed out.

Thanks,

askmeaboutlo0m:
There's some cool stuff in here, but probably too large for me to review currently. I'll look over it though and include the relevant bits into the `dancepile-2.2` branch, which is slated to become Drawpile 2.2 instead of `dev-2.2`.

askmeaboutlo0m:
I've been maintaining [my fork of Drawpile](https://github.com/askmeaboutlo0m/Drawpile/blob/dancepile-2.2/doc/changes-from-drawpile.md) since last year and have now taken over as maintainer of Drawpile proper. Reconciling the changes will probably involve cherry-picking the stuff that's compatible, there's definitely a lot of work done that can be applied, either as-is or with some changes. It's just anything involving the Rust part that's become moot, since the whole engine has been replaced.

csnover:
Fortunately, I haven’t made any changes to the Rust engine other than to fix some minor bugs and linting issues, and some preliminary work so that the paint engine can be restarted if the thread panics instead of forcing the user to quit and restart the whole application, so there isn’t any great loss in terms of anything I did.

The bulk of the changes are build system and improving UI/UX. Since similar changes were made on your branch reconciliation may be a challenge but since upstream isn’t dead now I am willing to make some effort to glue everything back together.

I would be interested to learn more at some point about the challenges that caused the Rust engine to be removed—for instance, I’ve been running with ASan and LTO OK, and whilst I haven’t compiled Rust to wasm personally, I know people who do this and might be able to help—but the #1 priority right now for me is to avoid the forks growing apart any more than they have already.

What will be best for you to talk about coordinating some of this?

askmeaboutlo0m:
Yeah, I've gathered that now by looking over it, but there's some duplicated stuff, like the vcpkg build, Qt6 support, updating the build system etc. I can do the work of reassembling that stuff, since I'm the one that broke upstream by subsumption. Unless you really want to take on those efforts.

This was probably one and a half years ago at this point, so things may have changed since then with Rust. I think ASAN just outright refused to work, I don't recall why exactly. LTO didn't work between C++ and Rust without injecting a wrapper script around clang and providing a dummy definition of `__rust_probestack` because of some bug in the LLVM gold linker, a known issue that had been open since years at that point. With regards to WebAssembly, the Emscripten target of Rust has/had bitrotted, the stated alternative being the raw wasm target. Which didn't support threading or have all the supporting bindings to WebGL and whatnot that Emscripten brings, and I wasn't about to write my own bootleg version of Emscripten when my interest was to write a web version of Drawpile. The removal then was simply a matter of not wanting to maintain two paint engines in parallel, with the Drawdance engine being the superior option in pretty much all aspects aside from the hypeness levels of its programming language.

You can contact me on Discord or Telegram. I don't wanna post the contact information here for fear of spammage, but you can find me in [the official Drawpile Discord server](https://discord.gg/M3yyMpC) or the announcement Discord server or Telegram channel at the top of [the Dancepile release pages](https://github.com/askmeaboutlo0m/Drawpile/releases/tag/2.2.0dancepiledev21).

csnover:
This is superseded by #1066.

PR #1061, opened 2023-01-14
Remove Krita X11 custom tablet support code

csnover:
This patch is based atop the previous PR #1060.

The custom tablet code was removed from Krita in upstream commit ec417d6c8217338cab2dbc6a13f9854c40786283, so presumably is not needed any more.


csnover:
This is superseded by #1066.

PR #1062, opened 2023-01-14
Replace clap with xflags

csnover:
This patch is based on top of PR #1060.

This change reduces the size of the drawpile-cli executable by over 4MiB and reduces compilation time on CI by 25%, in exchange for slightly less pretty (but just as useful) help output.


csnover:
This is superseded by this code not existing any more right now.

PR #1069, opened 2023-04-13
Add checks in Windows installer for OS prerequisites

csnover:
It is unnecessary to package vc_redist in most cases because it is usually already installed (and Windows Update apparently keeps it up-to-date as of 2019), so including that executable is just wasting disk space and bandwidth for most people.

Trying to download it on-demand in WiX Toolset is possible but burdensome, and would require either using `<CustomAction>` to hack in an on-demand download, or the ‘blessed’ way of creating a second separate bundle installer.

Since this problem is really quite rare, this commit just sets a precondition to make sure that a compatible version of Windows and a compatible version of vc_redist are installed, and prompts the user to manually download and run the installer if not.


askmeaboutlo0m:
Okay it did not understand that I merged this. It is merged though, d5d951be391956d1c74e7ab091bdc5561082a8a0

PR #1070, opened 2023-04-14
Make Windows Work

askmeaboutlo0m:
Makes KisTablet work again, now without needing a restart to apply its settings and without having to patch Qt. Instead, we install an event filter that will throw all of Qt's tablet events on the floor when KisTablet is enabled. We can distinguish them because Qt events are `spontaneous` and KisTablet ones are not.

Extends the tablet tester a little bit, since I needed it to check if it was working right.

Also fixes the problem of the brush database not getting created on first boot (Qt for some reason thought the dirname was `.` so the directory was never created), fixes resource inclusion (could use verification on OSX) and pulls in a fix for window titles that was merged into 2.1 but not 2.2.


cmake/AddResources.cmake
```diff
@@ -61,6 +68,14 @@ function(add_resources target)
 				"${android_dir}/assets/${file}"
 				SYMBOLIC
 			)
+		# Assets will already be installed by MACOSX_PACKAGE_LOCATION on macOS
+		elseif(NOT APPLE)
+			# No trailing slash is required or else it will strip the last path
+			# of the directory
+			install(FILES "assets/${file}"
+				DESTINATION "${INSTALL_APPDATADIR}${file_dir}"
+				COMPONENT "${target}"
```
csnover, line 77: Not directly related but probably should just delete the `COMPONENT` from everywhere; that was supposed to make it possible for the MSI installer to have choosable options for the client, the stand-alone server, and the command-line tools, but I couldn’t get CPack to correctly generate a non-monolithic installer (IIRC it just ran the `install(CODE)` whenever it felt like it no matter what I did) so now these are all just superfluous.

src/desktop/bundled/kis_tablet/kis_tablet_support_win.cpp
```diff
@@ -141,23 +141,37 @@ static inline bool isMouseEventType(QEvent::Type t)
...
                          delete QTAB;
                          QTAB = QWindowsTabletSupport::create();
                      });
 }
 
+void KisTabletSupportWin::quit()
+{
+    QObject::disconnect(conn);
+    conn = QMetaObject::Connection{};
+    delete QTAB;
```
csnover, line 171: `std::unique_ptr`?

src/desktop/dialogs/settingsdialog.cpp
```diff
@@ -124,13 +103,23 @@ SettingsDialog::SettingsDialog(QWidget *parent)
...
+	m_ui->tabletInputCombo->addItem(
+		tr("KisTablet Wintab"), int(tabletinput::Mode::KIS_TABLET_WINTAB));
+	m_ui->tabletInputCombo->addItem(
+		tr("KisTablet Wintab with Relative Pen Mode Hack"),
+		int(tabletinput::Mode::KIS_TABLET_WINTAB_RELATIVE_PEN_HACK));
 #	endif
-#	if defined(INPUT_NO_RESTART_REQUIRED)
-	m_ui->tabletInputRequiresRestartLabel->hide();
+#	if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+	m_ui->tabletInputCombo->addItem(
```
csnover, line 117: One of these Qt6 options must be the equivalent of `Mode::QT5` so I would just have one enumeration value for both of them (like `Default`).

src/desktop/tabletinput.cpp
```diff
@@ -18,73 +19,223 @@ namespace tabletinput {
...
+		switch(event->type()) {
+		case QEvent::TabletEnterProximity:
+		case QEvent::TabletLeaveProximity:
+		case QEvent::TabletMove:
+		case QEvent::TabletPress:
+		case QEvent::TabletRelease:
+			// KisTablet uses QApplication::sendEvent, which means its events
+			// are never set to be spontaneous. We use that to filter out Qt's
+			// own tablet events, which in turn are always spontaneous.
+			if(event->spontaneous()) {
```
csnover, line 47: I’m not saying that anything needs to change if this works, but I am curious if there is any particular reason why the KisTablet events shouldn’t be just using their own registered custom event type and then all of the Qt tablet events are simply ignored when KisTablet is on? Using the same event types provides better isolation between the consumer and producer, but code is already doing stuff based on KisTablet ifdefs anyway, so it doesn’t seem like there’s much practical benefit but an increased risk of future breakage. Is this true?

src/desktop/tabletinput.cpp
```diff
@@ -18,73 +19,223 @@ namespace tabletinput {
...
 }
+#	endif
 
-void update(const QSettings &cfg)
+#	if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+static void enableQt6TabletInput(bool wintab)
 {
-#if defined(Q_OS_WIN) && QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
 	// See by qtbase tests/manual/qtabletevent/regular_widgets/main.cpp
 	using QWindowsApplication = QNativeInterface::Private::QWindowsApplication;
```
csnover, line 175: I think this should be using `#include <QWindowsApplication>` and calling `QGuiApplication::nativeInterface`, they just failed to finish documentation to access the native interface: https://bugreports.qt.io/browse/QTBUG-100077

src/desktop/tabletinput.h
```diff
@@ -9,12 +9,44 @@ class QSettings;
 
 namespace tabletinput {
 
-void init(QApplication *app, const QSettings &cfg);
+#if defined(Q_OS_WIN)
+
+enum class Mode : int {
+	UNINITIALIZED,
```
csnover, line 15: The style for enum members is `CamelCase` not `UPPER_SNAKE`?

src/desktop/tabletinput.cpp
askmeaboutlo0m: Parts of Qt react to tablet events and if the regular Qt tablet input is broken for your device, turning on KisTablet would only fix the canvas view and the tablet tester, while the rest of the UI would still receive broken events. Qt also does mouse event synthesis if you don't accept tablet events (I think that's how it works anyway), so we couldn't just switch over to reading the other kind of event in KisTablet mode, we'd still have to handle the Qt tablet events to explicitly throw them on the floor. So just blocking the Qt tablet events and producing our own seems more stable.

src/desktop/dialogs/settingsdialog.cpp
askmeaboutlo0m: Qt5 mode is an undetermined and as far as I know unqueriable superposition of Windows Ink, Wintab and a big wad of nothing, depending on Windows version, Qt version and compile flags. Which doesn't match either Qt6 option enough for me to want to conjoin them.

.github/workflows/main.yml
```diff
@@ -258,7 +267,7 @@ jobs:
         uses: actions/upload-artifact@v3
         if: matrix.packager && startsWith(github.ref, 'refs/tags/') != 'true'
         with:
-          name: ${{ matrix.cross_os || runner.os }} binaries
+          name: Drawpile-${{ matrix.cross_os || runner.os }}-Qt${{ matrix.qt }}
```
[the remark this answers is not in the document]
askmeaboutlo0m: Putting Drawpile- in front of it just makes it easier to find the file when it gets dumped into a downloads directory, putting the Qt version after it is there so that it's clear which job it belongs to, since it's not immediately obvious that only the Qt5 jobs create release artifacts.

src/desktop/tabletinput.h
askmeaboutlo0m: Oops, fixed.

src/desktop/bundled/kis_tablet/kis_tablet_support_win.cpp
askmeaboutlo0m: Used that here and also in similar situations in tabletinput.cpp.

src/desktop/tabletinput.cpp
askmeaboutlo0m: Trying to get this working, it seems like `QWindowsApplication` doesn't work that way, it's a platform interface or something not a native interface. The header doesn't exist anyway and unlike `QNativeInterface::QX11Application`, there appears to be no `QNativeInterface::QWindowsApplication`.

.github/workflows/main.yml
```diff
@@ -147,6 +147,15 @@ jobs:
             packager: >
               cp .github/deps/other/bin/qt*.dll .github/deps/qt/bin &&
               cpack --verbose --config build/CPackConfig.cmake -C Release
+
+          # Windows Qt 6
+          - os: windows-latest
+            qt: 6.4.2
+            # KisTablet crashes on Qt6. Unlike Qt5, it does allow for switching
```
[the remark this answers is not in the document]
askmeaboutlo0m: I don't see a difference in how it works in your branch anyway, but it could be that the problem occurs because it fails to uninstall the KisTablet handlers correctly or something. I'd leave it like that for now, maybe a Windows dev wants to look into it or I'll decide to set up a Windows build environment in Qt6 sometime and try to debug it there.

cmake/AddResources.cmake
```diff
@@ -53,6 +53,13 @@ function(add_resources target)
 	foreach(file IN LISTS ARG_FILES)
 		target_sources(${target} PRIVATE "assets/${file}")
 		set_property(TARGET ${target} APPEND PROPERTY RESOURCE "assets/${file}")
+		get_filename_component(file_dir "${file}" DIRECTORY)
+		if(NOT "${file_dir}" STREQUAL "")
+			set(file_dir "/$file_dir")
+		endif()
+		set_source_files_properties("assets/${file}"
```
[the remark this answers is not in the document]
askmeaboutlo0m: Removed this bit then. I think that was the correct thing to do?

cmake/AddResources.cmake
```diff
@@ -61,6 +68,14 @@ function(add_resources target)
 				"${android_dir}/assets/${file}"
 				SYMBOLIC
 			)
+		# Assets will already be installed by MACOSX_PACKAGE_LOCATION on macOS
+		elseif(NOT APPLE)
```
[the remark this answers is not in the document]
askmeaboutlo0m: Removed it. I don't think the brush database crashed, but color palettes did if they were missing.

cmake/AddResources.cmake
askmeaboutlo0m: Removed 'em.

src/desktop/tabletinput.cpp
csnover: Argh, sorry. `QNativeInterface::Private::QWindowsApplication` is still the only one for now. It does look like it should be accessible from `QGuiApplication::nativeInterface` at least, if [this code](https://github.com/wangwenx190/framelesshelper/blob/3b8571401fab94e9eb7bd3dce3b924e08a2784cf/src/core/utils_win.cpp#L1623) and [this code](https://github.com/rgriebl/brickstore/blob/b509923c92a7e3e02a3c8ec0177530ef7665d900/src/common/systeminfo.cpp#L129) are believable, if it matters at all to get rid of one private API.

src/desktop/tabletinput.cpp
csnover: OK. This is all just for future reference, since I bothered to look:

`Qt::AA_SynthesizeMouseForUnhandledTabletEvents` appears to control Qt’s automatic synthesis if the platform itself doesn’t already generate its own synthetic mouse events. Things I have noticed that it does with tablet events is (1) update the QGuiApplication modifier buttons state, (2) track what window is being targeted for events, (3) forward and/or ignore events in the same way as mouse events.

cmake/AddResources.cmake
```diff
@@ -45,14 +45,20 @@ function(add_resources target)
...
 			)
 		endif()
 	endforeach()
 
 	foreach(file IN LISTS ARG_FILES)
 		target_sources(${target} PRIVATE "assets/${file}")
 		set_property(TARGET ${target} APPEND PROPERTY RESOURCE "assets/${file}")
+		get_filename_component(file_dir "${file}" DIRECTORY)
+		if(NOT "${file_dir}" STREQUAL "")
+			set(file_dir "/$file_dir")
```
csnover, line 57: I am blind and did not notice this is not a variable until now, it is replacing `file_dir` with literally `/$file_dir` (should be `/${file_dir}`). This if-branch shouldn’t be needed at all though.

cmake/AddResources.cmake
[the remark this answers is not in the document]
csnover: Adding a file to the target’s `RESOURCE` property, and having a source file with the `MACOSX_PACKAGE_LOCATION` property, are mutually exclusive choices. If a file is in the `RESOURCE` list, `MACOSX_PACKAGE_LOCATION` is ignored and the file is dumped into the Resources directory.

`install(TARGET … RESOURCE DESTINATION)` (when it was working?!) is the same: the `RESOURCE` property is a flat list of files, and they all just get dumped into the destination as if you ran `install(FILES)` on that list.

So if this needs to support selecting individual files from subdirectories of the assets directory (which right now I don’t think it does?), then either the `RESOURCE` target property has to be abandoned totally and `MACOSX_PACKAGE_LOCATION` is used exclusively, or else `RESOURCE` has to be used only for files with no subdirectory and `MACOSX_PACKAGE_LOCATION` has to be used only for files that have subdirectories.

Hopefully this is clear now, because it sure took me about a week the first time to figure out WTF this thing is doing. :-)

src/desktop/dialogs/settingsdialog.cpp
csnover: However, the fallback code already makes it equivalent to `Qt6Winink`?

cmake/AddResources.cmake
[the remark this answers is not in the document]
askmeaboutlo0m: Okay, that makes reasonable amounts of sense, as far as CMake goes anyway. It seems there's a whole lot of ways to do installation, and they all seem to be broken in different ways. I should probably also pay attention to what I'm doing, cause I obviously just went and un-fixed top-level files not getting installed. Reverted that.

src/desktop/dialogs/settingsdialog.cpp
askmeaboutlo0m: It's gotta fall back to *something*, I picked the option that's less broken. Wintab in Qt6 causes menus to only react to pen inputs very sporadically, most of the time they just ignore you.

Conflating Qt6Winink and Qt5MysteryInput into QtDefault would still be possible, but I don't think it would make anything shorter or clearer, all it changes is that that one fallback is implicit. The UI still needs to call them different names and the settings code still has to do different things depending on which Qt version it's under.

src/desktop/tabletinput.cpp
askmeaboutlo0m: Ah, that would explain why we get such different behavior on Windows and Linux, maybe one of those platforms generates synthetic events on its own. The tablet tester on Windows always shows both a tablet trail and a mouse trail, while Linux only shows a tablet trail (for me anyway.)

src/desktop/tabletinput.cpp
askmeaboutlo0m: Would definitely be nicer. Changed it to that effect and will give it another try to see if it still works that way.

src/desktop/tabletinput.cpp
```diff
@@ -18,73 +20,222 @@ namespace tabletinput {
...
+	static constexpr Mode qt6Wintab = Mode::Qt5;
+#	endif
+
+#	ifdef HAVE_KIS_TABLET
+	static constexpr Mode kisTabletWinink = Mode::KisTabletWinink;
+	static constexpr Mode kisTabletWintab = Mode::KisTabletWintab;
+	static constexpr Mode kisTabletWintabRelativePenHack =
+		Mode::KisTabletWintabRelativePenHack;
+#	else
+	static Mode kisTabletWinink = qt6Winink;
```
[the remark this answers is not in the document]
askmeaboutlo0m: It's doing nothing useful other than maybe waste some space in the bss segment. I blame programming in grayed-out ifdefs for not noticing that it's nonsense.

src/desktop/tabletinput.cpp
```diff
@@ -18,73 +20,222 @@ namespace tabletinput {
...
+		return kisTabletWintab;
+	case int(Mode::KisTabletWintabRelativePenHack):
+		return kisTabletWintabRelativePenHack;
+	case int(Mode::Qt5):
+		return qt5;
+	case int(Mode::Qt6Winink):
+		return qt6Winink;
+	case int(Mode::Qt6Wintab):
+		return qt6Wintab;
+	default:
```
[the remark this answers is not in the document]
askmeaboutlo0m: Yeah, that's nicer. Changed to that effect.

cmake/AddResources.cmake
askmeaboutlo0m: Yeah evidently it doesn't end up getting called, so I'll just delete it.

src/desktop/tabletinput.cpp
```diff
@@ -2,89 +2,232 @@
...
+	constexpr Mode qt6Winink = Mode::Qt6Winink;
+	constexpr Mode qt6Wintab = Mode::Qt6Wintab;
+#	else
+	constexpr Mode qt5 = Mode::Qt5;
+	constexpr Mode qt6Winink = Mode::Qt5;
+	constexpr Mode qt6Wintab = Mode::Qt5;
+#	endif
+
+	Mode mode = getModeFromSettings(cfg);
+	switch(getModeFromSettings(cfg)) {
```
csnover, line 92: making the same call twice here

src/libclient/brushes/brushpresetmodel.cpp
```diff
@@ -18,6 +18,7 @@
 #include <QSqlDatabase>
 #include <QSqlError>
 #include <QSqlQuery>
+#include <QThread>
```
csnover, line 21: iwyu or accident?

cmake/AddResources.cmake
```diff
@@ -61,6 +64,14 @@ function(add_resources target)
 				"${android_dir}/assets/${file}"
 				SYMBOLIC
 			)
+		# Assets will already be installed by MACOSX_PACKAGE_LOCATION on macOS
+		elseif(NOT APPLE)
+			# No trailing slash is required or else it will strip the last path
+			# of the directory
+			install(FILES "assets/${file}"
+				DESTINATION "${INSTALL_APPDATADIR}${file_dir}"
+				COMPONENT "${target}"
```
csnover, line 73: Missed a `COMPONENT` to delete here

cmake/AddResources.cmake
```diff
@@ -45,14 +45,17 @@ function(add_resources target)
...
 			)
 		endif()
 	endforeach()
 
 	foreach(file IN LISTS ARG_FILES)
 		target_sources(${target} PRIVATE "assets/${file}")
 		set_property(TARGET ${target} APPEND PROPERTY RESOURCE "assets/${file}")
+		get_filename_component(file_dir "${file}" DIRECTORY)
+		set_source_files_properties("assets/${file}"
+			PROPERTIES MACOSX_PACKAGE_LOCATION "Resources${file_dir}"
```
csnover, line 57: This still needs to be using either `set_property(TARGET … APPEND PROPERTY RESOURCE <filename>)` ***or*** `set_source_file_properties(<filename> PROPERTIES MACOSX_PACKAGE_LOCATION)`, not both at the same time.

If `TARGET… RESOURCE` is used instead, that does not support subdirectories, and so there should not be any attempt to do stuff with `${file_dir}` because that won’t work.

cmake/AddResources.cmake
```diff
@@ -61,6 +64,14 @@ function(add_resources target)
 				"${android_dir}/assets/${file}"
 				SYMBOLIC
 			)
+		# Assets will already be installed by MACOSX_PACKAGE_LOCATION on macOS
+		elseif(NOT APPLE)
+			# No trailing slash is required or else it will strip the last path
+			# of the directory
```
csnover, line 70: This comment doesn’t actually apply to `install(FILES)`, since the problem of trailing slashes has to do with the source of an `install(DIRECTORY)`.

src/desktop/tabletinput.cpp
askmeaboutlo0m: Corrected to use the variable as it should.

src/libclient/brushes/brushpresetmodel.cpp
askmeaboutlo0m: Leftover attempt of adding retries in addition to error messages, since the error messages made it clear that the problem was that Qt's dirname wasn't working as expected. Removed it.

cmake/AddResources.cmake
askmeaboutlo0m: Purged the comment.

cmake/AddResources.cmake
askmeaboutlo0m: I'm starting to think there's something wrong with my vscode, it keeps opening wrong files and not finding the stuff I grep for. Removed it.

cmake/AddResources.cmake
askmeaboutlo0m: Another thing I could swear I grepped through and removed. Deleted the TARGET thing.

csnover (approved):
I don’t see any new thing to provide useful feedback on so it seems fine to me to land. There will be some lingering issue I need to resolve with the change to how the windows driver setting is recorded in the configuration file but that is fine, the new settings component is ostensibly designed so things like this can be papered over.

askmeaboutlo0m:
Excellent, thank you!

PR #1071, opened 2023-04-15
Implement dynamic theme and dark mode switching

csnover:
This patch makes it possible to switch themes without restarting and also correctly responds to OS-level theme changes so the application does not need to be restarted if the OS theme changes.

System light and dark mode theme options are added; on macOS 10.14+ these are the native OS light/dark palettes, and on other OS it falls back to the Krita Bright or Fusion Dark palettes using the system style.

This patch also fixes the appearance of the rotation widget in dark mode, and the appearance of `GroupedToolButton` on all OS, since that widget was doing OS-specific stylesheet stuff which was wrong and fixing that meant just fixing the widget’s bad painting in general.


src/desktop/widgets/groupedtoolbutton.cpp
```diff
@@ -63,78 +106,69 @@ void GroupedToolButton::setColorSwatch(const QColor &c)
...
+		const auto *parentStyle = parentWidget()
+			? parentWidget()->style()
+			: QApplication::style();
+
+		// `QProxyStyle` takes ownership of the base style, but `setStyle` does
+		// not take ownership of whatever style is passed to it, so we need to
+		// track and delete our own proxy styles (but not whatever may be
+		// returned that is inherited since that is owned by something else)
+		auto *oldStyle = qobject_cast<GroupedToolButtonStyle *>(style());
+		setStyle(new GroupedToolButtonStyle(parentStyle->name()));
```
askmeaboutlo0m, line 162: This leaks in exactly one case, I guess because Qt isn't calling the destructor or something. Setting the destructor `virtual` doesn't help, but using Qt-sanctioned ownership does. Qt5 also doesn't have `QStyle::name`, but `QObject.:objectName` works for me (probably something to move into qguicompat).

```suggestion
#	if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
		QString styleName = parentStyle->name();
#else
		QString styleName = parentStyle->objectName();
#endif
		auto *newStyle = new GroupedToolButtonStyle(styleName);
		newStyle->setParent(this);
```

And also removing the explicit destructor, since it's no longer needed when Qt is doing it.

csnover: I’ve revised this so it should be OK now. I ran a leak check and this didn’t look like it was leaking any more, but about a dozen things in QMacStyle are 🙃

csnover:
This should be OK to go now.

askmeaboutlo0m (approved):
The angle gauge uses wrong colors and this is missing the hot dog stand theme. Despite the latter being a major shortcoming, it doesn't need to be fixed in this PR, so I'll merge it in.

PR #1072, opened 2023-04-16
Use order by LOWER(name) in tag picking query

askmeaboutlo0m:
Just a small fix, this made the choices from the tags dropdown not match up with what actually ended up being selected when choosing them.


csnover (approved):
LGTM

askmeaboutlo0m:
This was merged as 1aaf83da61a14b20f4a24c00c9bdaf644b0466e5, Github just ain't that smart.

PR #1073, opened 2023-04-17
Ensure svg iconengine is included in packaged AppImage

csnover:
Without this, icon lookups fail confusingly by hunting for PNGs and not SVGs within the entire universe of XDG icon directories.

Meta commentary, I don’t feel like the review requirement protected branch rule is a good long-term rule since it reduces velocity, particularly in cases of trivial fixes like this.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -72,7 +72,11 @@ jobs:
             # This causes the AppImage to be generated, instead of just creating
             # the portable tree, because there seems to be no way to separate
             # these steps with linuxdeploy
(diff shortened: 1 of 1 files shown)


askmeaboutlo0m:
Yup, that works good. The branch config definitely isn't right like this and sure isn't what I intended, but in the current whirlwind mode it'll do. There's been some mess with who has access to the project recently, with someone getting their access pulled, so there may be some other wonk to clean up anyway, I'll get to that eventually so that it's actually usable once we get to cruising speeds again.

PR #1074, opened 2023-04-18
Continuous Release

askmeaboutlo0m:
Generates a release named `continuous` under a tag with the same name for every successful build on main.

Example run: <https://github.com/drawpile/Drawpile/actions/runs/4734393156>

Example release: <https://github.com/drawpile/Drawpile/releases/tag/continuoustest>

Will probably wait for #1070 to get merged so that the release it generates will actually work on Windows.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,6 +3,12 @@ name: Build and Test
 on:
   pull_request:
   push:
+    branches:
+      - '**'
+    tags-ignore:
+      # This tag is (re)created when a build on main succeeds, it'd be
+      # pointless to trigger yet another build in response to it.
+      - continuous
   schedule:
     - cron: '18 11 * * *'
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -281,28 +281,10 @@ jobs:
 
       - name: Upload artifact
         uses: actions/upload-artifact@v3
-        if: matrix.packager && startsWith(github.ref, 'refs/tags/') != 'true'
+        if: matrix.packager
         with:
           name: Drawpile-${{ matrix.cross_os || runner.os }}-Qt${{ matrix.qt }}
           path: |
-            Drawpile-*.AppImage
-            Drawpile-*.apk
-            Drawpile-*.dmg
-            Drawpile-*.msi
-            Drawpile-*.zip
-
-      - name: Collect release notes
-        run: awk -v RS='' '/^[[:digit:]]{4}-[[:digit:]]{2}-[[:digit:]]{2} Version ${{ github.ref_name }}/,/^[[:digit:]]/' ChangeLog | tail '+2' > ChangeLog-release
-        if: matrix.packager && startsWith(github.ref, 'refs/tags/') == 'true'
-
-      - name: Upload release
-        uses: softprops/action-gh-release@v1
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -309,3 +315,34 @@ jobs:
...
+      - name: Dump directory contents
+        run: ls -alFR
+
+      - name: Create continuous release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TARGET_COMMIT: ${{ github.sha }}
+          RELEASE_NAME: continuous
+          RELEASE_DESCRIPTION: >
+            This is an automatic, continuous development release based on the
```
csnover, line 336: > This release is an automatically generated snapshot of the current state of development. It is continuously updated with work-in-progress changes that may be broken, incomplete, or incompatible with other versions of Drawpile.

.github/workflows/main.yml
```diff
@@ -309,3 +315,34 @@ jobs:
...
+
+  continuousrelease:
+    needs: test
+    runs-on: ubuntu-latest
+    if: success() && github.ref == format('refs/heads/{0}', github.event.repository.default_branch)
+    steps:
+      - name: Download artifacts
+        uses: actions/download-artifact@v3
+
+      - name: Dump directory contents
```
csnover, line 327: Left in from debugging?

pkg/make-github-release.sh
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
askmeaboutlo0m: Deleting the release through the Github CLI sure sounds easier, I'll give that a try. I don't like the release upload action too much though because it doesn't label the artifacts properly, you just get raw file name gibberish that a mortal can't tell apart. That too can be replaced with a `gh release create ...` though apparently.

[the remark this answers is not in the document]
askmeaboutlo0m: My experience has been that a significant number of people will see a pile of links that are "the same" followed by two links that say "code". Okay, this is clearly techy tech stuff and there's even coding involved somehow, better ask which buttons they're supposed to press.

Even if they manage to spot that the file extensions are distinct, they may not recognize what's what from that alone. I don't have first-hand painful experiences with that, since I've only been uploading zips, but I could very well imagine that ".msi" is not a particularly well-known format (most installers seem to be .exe) and translating ".zip" to "portable Windows version" is quite a leap, especially given that right below it there's another zip option for source code.

Replacing the bulk of the script with two tiny Github CLI calls has worked anyway, I think that solution is reasonably short while giving a result with pretty names.

.github/workflows/main.yml
askmeaboutlo0m: Reworded.

askmeaboutlo0m:
Now works for both the continuous release as well as the regular, tagged releases, so that they all get pretty artifact labels.

csnover (approved):
Something about the asset names still bugs me for some reason but I think it is just subjective, nothing to prevent this from going.

askmeaboutlo0m:
Awaiting #1076 because the current continuous release already contains the animation rework since I had some folks test it and I don't wanna clobber that.

PR #1075, opened 2023-04-25
Fix unpredictable tabletinput segfault on exit

askmeaboutlo0m:
By removing the use of unique_ptr so that we don't run into the Reverse Static Initialization Order Fiasco. Globals are destructed in effectively random order, so it may decide to destroy the QApplication before destroying the event filter, causing a crash because the event filter is trying to deregister itself from an app that no longer exists.


askmeaboutlo0m:
Cow says looks good, so merging this.

PR #1077, opened 2023-04-25
Various fixes on windows

cow54787:
- Fix clang on windows
- Fix missing header guards
- Fix missing #include
- Fix symbol conflict with Qt6
- Copy assets to the build directory
- Unity build option


askmeaboutlo0m:
Looks good, merged as d1371d36f61665465ce06156d741b1e54912c608 because Github doesn't understand rebases. Unity build gets a bit wonky when tests get involved, but that's expected, since the codebase isn't really set up for it.

PR #1079, opened 2023-04-27
Android Repair

askmeaboutlo0m:
Makes Android work fully again, bringing back icons, making ZIP stuff work, fixing the chat input and producing a signed release mode APK where appropriate.


askmeaboutlo0m:
Build failures are due to cache incompatibility from new Qt patches and the KF5Archive dependency.

.github/scripts/patches/qandroidassetsfileenginehandler_dot_slash.diff
```diff
@@ -0,0 +1,34 @@
+--- a/src/plugins/platforms/android/qandroidassetsfileenginehandler.cpp	2023-04-26 08:39:21.720887254 +0200
++++ b/src/plugins/platforms/android/qandroidassetsfileenginehandler.cpp	2023-04-26 08:39:24.477553939 +0200
+@@ -55,6 +55,7 @@
+     if (file.startsWith(assetsPrefix))
+         file.remove(0, prefixSize);
+     file.replace(QLatin1String("//"), QLatin1String("/"));
++    file.replace(QLatin1String("/."), QLatin1String(""));
```
csnover, line 7: These changes would break any files that start with a dot. In lieu of a more accurate normaliser that actually loops over each path fragment to apply "", ".." and ".", each of these sections will need to do `if (file.startsWith("./")) file.remove(0, 2);`, `if (file.endsWith("/.")) file.chop(2);`, and `file.replace("/./", "/");`.

.github/actions/build-deps/action.yml
```diff
@@ -48,9 +48,13 @@ inputs:
     default: c6f0b66318f8da6917fb4681103f7303b1836194
     type: string
   libzip:
-    description: Version of libzip to build
+    description: Version of libzip to build (Qt6 only)
```
csnover, line 51: libzip works with Qt6 on Android?

src/desktop/CMakeLists.txt
```diff
@@ -363,6 +363,8 @@ else()
 	# Android can neither deal with exporting multiple files nor can it spawn an
```
csnover, line 363: Outdated/incomplete comment now

src/desktop/CMakeLists.txt
```diff
@@ -346,6 +344,8 @@ if(ANDROID)
 	# other up with no way to switch between them, so we don't allow that there.
```
csnover, line 344: Outdated/incomplete comment now

src/desktop/chat/chatlineeditandroid.cpp
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
csnover, line 1: Why create a whole separate TU for this? The idiomatic way to handle message send on every Android messaging app I have ever used (and some desktop apps too) with multi-line chat is to have a send button; can we just add one of those? (It would be a good idea to have a send button in any case since it is conventional.)

src/desktop/chat/chatwidget.h
```diff
@@ -78,6 +78,13 @@ private slots:
 	void expandRequested();
 
 private:
+#ifdef Q_OS_ANDROID
```
csnover, line 81: Why not just `#ifdef` out the code directly? Is there some plan for this to apply elsewhere later?

src/desktop/chat/chatwidget.h
```diff
@@ -79,9 +79,13 @@ private slots:
 
 private:
 #ifdef Q_OS_ANDROID
+	// On Android, image resources don't work, leading to avatars not showing.
```
csnover, line 82: Why don’t they work?

.github/scripts/patches/qandroidassetsfileenginehandler_dot_slash.diff
csnover: Actually, both `QDir` and `QUrl` have a path normaliser so I am not sure why the Android QPA is doing this half-baked thing since it could access the private APIs (`qt_normalizePathSegments`).

.github/scripts/build-qt.cmake
```diff
@@ -220,6 +220,7 @@ if(BASE)
 				patches/cast_types_for_egl_x11_test.diff
 				patches/qmake_android_ltcg.diff
 				patches/qtbug-112504.diff
+				patches/qandroidassetsfileenginehandler_dot_slash.diff
```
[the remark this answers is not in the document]
askmeaboutlo0m: No, I still gotta set up a build environment so I can actually provide something sensible to reproduce. Same with the Qt6 menu bugs. But I got it written down.

.github/scripts/patches/qandroidassetsfileenginehandler_dot_slash.diff
askmeaboutlo0m: Since these are asset paths, no files are gonna start with a dot. But I'll take a look if that normalize function will act properly.

.github/actions/build-deps/action.yml
askmeaboutlo0m: Not properly, but it works with Qt6 on desktop. KF6 KArchive will presumably work on Android though, but I don't think it's out yet still.

src/desktop/chat/chatwidget.h
askmeaboutlo0m: The ifdefs just added a lot of clutter in headers, sources of both this file and the signal connections made to it, to the point where I got myself confused a few times. So it's just less of a mess, but it could just as well be ifdef'd out directly if you think that would make more sense.

src/desktop/chat/chatwidget.h
askmeaboutlo0m: Avatars in chat use a special kind of URL and apparently Android has some kind of URL restrictions that make it not work with those. I didn't look too deep into after prior dives into Qt's Android support, I just turned them off instead.

.github/scripts/patches/qandroidassetsfileenginehandler_dot_slash.diff
askmeaboutlo0m: Linking fails when attempting to use that function, even though it's linking with the module itself, so I guess it's not an exported function that's usable without also getting into messing with that. I don't particularly wanna do that to solve a nonexistent issue, but it's something to suggest when reporting it to Qt.

src/desktop/CMakeLists.txt
askmeaboutlo0m: Updated the comments. The main window one got roped into it too because it's all just consolidated now.

PR #1084, opened 2023-05-07
Rewrite settings code and UI (and minor miscellaneous fixes)

csnover:
This patch rewrites all code related to persistent settings in the Drawpile client.

(1) Replaces direct use of QSettings with an abstraction layer.

The new abstraction layer includes these improvements:

1. It is strongly typed, so callers no longer have to specify the type of the data they expect.

2. It brings together all the settings into a single table so it is easy to review and define new settings, and there is never a chance of keys being typoed or not using the correct default value if a setting does not exist.

3. It uses signals for changes to settings, so components can easily dynamically update in response to settings changes.

4. It has per-setting versioning mechanism so users can run different versions simultaneously without losing or corrupting settings.
   
   This is currently used only to support updates from 2.1 to 2.2 by detecting whether 2.1 settings exist or not and only writing old settings if they already existed.

6. Settings can be non-persistent or derived from other persistent data, though this mechanism is not really used at the moment.

7. It provides data bindings for common widgets so it is not necessary to `connect` the same signals over and over again for every combo box, check box, etc.

8. It provides a capability to defer committing changes to storage (for e.g. OK/Apply/Cancel). However, it turns out that pretty much every desktop environment now recommends to apply settings immediately, so this function is not used right now.

Some settings that were not persisted to QSettings like the certificate store are now using a correct Qt data model, instead of before where there was some pretty sus code using QListWidget and doing data management from the UI code.

(2) Rewrites the preferences window.

The settings UI has been completely redesigned. It now follows OS HIGs more closely and code has been split up by page instead of being a giant Designer UI monolith. The layout components have been customised to fix various form layout bugs and limitations in the built-in Qt layouts.

Tablet tester has been put directly in the input settings, though at the moment its input does not go through the component that smooths and applies pressure curves, so this needs to be fixed very soon to avoid confusing users when changing those settings does not change the tester output.

Language settings are now pulled directly from CMake instead of scanning the filesystem, so this is marginally quicker and also should be less error-prone since now there is only one place where i18n files are read instead of multiple places.

(3) Improves settings persistence.

Previously there were many settings, like the main window layout, that did not save unless the application went through the normal shutdown process. Settings are now constantly updated during use so if there is a crash or someone quits using Ctrl-C all the time, Drawpile will recover correctly instead of losing all changes since the last time it exited normally.

(4) Improvements to content filtering.

The “parental control” feature has been redesigned around the ethos of empowering individuals to make choices they feel are right for them.

1. Users can now ignore advisory tags from the server when deciding what content to filter, opting instead to maintain their own list of keywords that are not safe for them.

2. The local list of unsafe keywords is now actually applied to incoming sessions; previously, the local list of keywords was only used to automatically set the advisory tag on sessions.

(5) UI fixes to some other components.

Some other components that just needed UI fixes generally related to list servers and avatars have been partially updated in this patch because it became too annoying to try to split those out since they also needed changes to support the new settings system or the improvements to content filtering.


src/desktop/dialogs/settingsdialog/contentfilter.cpp
```diff
@@ -0,0 +1,195 @@
...
+	auto *hide = new QCheckBox(tr("Hide settings when locked"));
+	lockLayout->addWidget(hide);
+	settings.bindContentFilterHideLocked(hide);
+
+	lockLayout->addStretch();
+
+	auto *hiddenPage = new QWidget;
+	auto *hiddenLayout = new QVBoxLayout(hiddenPage);
+	hiddenLayout->setContentsMargins(0, 0, 0, 0);
+	hiddenLayout->addWidget(new QLabel(tr("Content filtering settings are currently locked.")), 0, Qt::AlignCenter);
```
askmeaboutlo0m, line 108: "Content filtering" makes it sound like we filter the content of sessions, which we sure don't. Parental controls seems like a fine and common name for them. Dito for all uses of this term.

askmeaboutlo0m (changes requested):
Nothing exciting found with the code, but the content filter verbiage needs changing.

Yet gotta try this out on Linux and Windows (for the tablet driver guff), so I might find some more things.

askmeaboutlo0m:
On Linux, the preferences sidebar looks a bit nebulous.
image

askmeaboutlo0m:
On Linux, when starting from a blank slate, the theme appears as a Fusion theme with system colors.
image

The preferences dialog purports it to be Windows instead of Fusion, which is lies.
image

It's supposed to be Fusion using Krita Dark by default on all platforms except on macOS. Doing it that way has virtually entirely removed any users nitpicking inane UI changes from Drawpile 2.1 because they are instead distracted by the sleekness of the dark mode.

When not starting from a blank slate, it seems to use the Fusion style, but reverts to system colors. I don't think it's necessary to add migration code to account for the few users that would be affected by this though, it can just clobber it with Krita Dark.

askmeaboutlo0m:
Oh yeah, the settings dialog lacks a button to close it. On systems where you don't have window decorations that makes it kinda hard to get rid of. On my window manager I have to hit a key combination, on Android you'd have to hit the "physical" back button.

askmeaboutlo0m:
The top checkbox in the parental controls ~~remains enabled when they're locked~~ exists. It shouldn't be an option, filtering out NSFM sessions is always done, keywords are additional measures against human error.

image

src/libclient/net/sessionlistingmodel.cpp
```diff
@@ -196,35 +229,60 @@ QVariant SessionListingModel::data(const QModelIndex &index, int role) const
...
-			return Qt::NoItemFlags;
+	if (isRootItem(index)) {
+		return Qt::ItemIsEnabled;
+	}
+
+	const auto &listing = m_listings.at(listingIndex(index));
+	if(listing.offline())
+		return Qt::ItemIsEnabled | Qt::ItemNeverHasChildren;
+
+	const auto &session = listing.sessions.at(index.row());
```
askmeaboutlo0m, line 263: Getting an index out of bounds right here. Adding a check leads to a busted server list where all items are disabled and toggling any of the filters takes an eternity to apply, so it seems like it's just going way off the deep end somewhere.

Reproduction setup: start with an empty set of list servers, Session → Join, type `pub.drawpile.net` into the input, click Add Server, accept the terms and conditions.

The error reproduces every time the session list is loaded from the list server.

askmeaboutlo0m: Hrm, I did a clean build now and it doesn't crash anymore, so maybe the moc just got messed up or something. So actually, the listings now don't crash.

src/libclient/net/sessionlistingmodel.cpp
```diff
@@ -196,35 +229,60 @@ QVariant SessionListingModel::data(const QModelIndex &index, int role) const
...
+	if(!session.protocol.isCurrent())
```
askmeaboutlo0m, line 264:
```suggestion
	if(!session.protocol.isCompatible())
```

Otherwise you can't join sessions in compatibility mode.

askmeaboutlo0m:
Qt5 fails to save all enums (video exporter format) and enum classes (theme colors, brush cursor types, tablet driver mode probably has the same issue), saying `QVariant::save: unable to save type`.

No attempts at `Q_ENUM` and `Q_ENUM_NS` help, but `Q_ENUM_NS` is probably a better solution for those enum classes declared at the namespace level than just declaring a metatype, since Qt seems to get a bit confused trying to print out nonexistent enum values.

Weirdly, it * does*  write stuff into the config file, but I guess it's garbage and definitely doesn't get loaded again.

Trying to change it to just use an `enum` and an `int` default value (because I guess that's how the type is determined?) instead of `enum class` doesn't work because I guess C++ treats enums as incompatible with integers in the parallel universe of template types. Changing them to bare `constexpr int`s works fine, of course.

askmeaboutlo0m:
Canvas scrollbars are always visible on startup, have to toggle them on and off to get rid of them.

src/libclient/net/sessionlistingmodel.cpp
csnover: This is fixed now and was due to an off-by-one error in the model caused by Qt’s use of inclusive ranges.

src/libclient/net/sessionlistingmodel.cpp
csnover: This is fixed now.

src/desktop/dialogs/joindialog.cpp
```diff
@@ -427,9 +415,9 @@ void JoinDialog::addListServer()
 	QString urlStr = m_ui->address->currentText();
 	QUrl url;
 	if(!urlStr.contains('/')) {
-		url = QUrl { "http://" + urlStr };
+		url = QUrl { "https://" + urlStr };
 	} else {
-		url = QUrl { "http://" + QUrl{urlStr}.host() };
+		url = QUrl { "https://" + QUrl{urlStr}.host() };
```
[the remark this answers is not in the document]
csnover: I have updated this so it should only default to HTTPS if a scheme is not explicitly specified, instead of the old code which had been always forcing to insecure HTTP and which I had only just updated to always force to HTTPS instead.

src/desktop/dialogs/joindialog.cpp
```diff
@@ -344,59 +336,55 @@ void JoinDialog::resolveRoomcode(const QString &roomcode, const QStringList &ser
...
 
-	const auto *listingHeader = m_ui->listing->header();
-
-	cfg.setValue("listsortcol",
-		(listingHeader->sortIndicatorSection() + 1) *
-		(listingHeader->sortIndicatorOrder() == Qt::AscendingOrder ? 1 : -1)
-	);
+	QStringList hosts;
+	auto max = settings.maxRecentFiles();
+	if (max) {
```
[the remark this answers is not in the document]
csnover: This change has been made.

src/desktop/tabletinput.cpp
```diff
@@ -218,24 +163,39 @@ void update(QApplication *app, const QSettings &cfg)
...
 	// Nothing to do on other platforms.
 	Q_UNUSED(app);
-	Q_UNUSED(cfg);
 #endif
 }
 
 QString current()
 {
-	return QString::fromUtf8(inputMode);
+	switch (currentMode) {
```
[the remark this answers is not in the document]
csnover: It sounds like that should probably be transmitting just the raw enumeration value instead of a full string. For now I have modified this change so this function now returns the translation source string literal instead of `QString`, and then translates in the About window.

FWIW, anything i18n-related I have been pretty lazy about it in this patch because I know all of those need to be touched again soon when the dynamic translations stuff gets reworked and added.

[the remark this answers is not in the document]
askmeaboutlo0m: Yeah, that would make more sense, now that that enum actually exists.

csnover:
> The preferences dialog purports it to be Windows instead of Fusion, which is lies.

This was caused by the default style having an empty string name, where `QStyleFactory` only returns named styles. It is fixed now by adding a separate “System” option for the empty string theme style.

> It's supposed to be Fusion using Krita Dark by default on all platforms except on macOS. Doing it that way has virtually entirely removed any users nitpicking inane UI changes from Drawpile 2.1 because they are instead distracted by the sleekness of the dark mode.

By defaulting to a non-system palette, users who require a high contrast colour scheme for accessibility reasons will find it difficult or impossible to use Drawpile because they will not be able to navigate through the preferences to switch the colour scheme back, so I would ask to please not do this. For nitpicking feedback, you can feel free to send anyone doing that in my direction if it would reduce the burden.

> When not starting from a blank slate, it seems to use the Fusion style, but reverts to system colors.

This is fixed now and was caused by an incomplete implementation of the settings migration code.

> Qt5 fails to save all enums (video exporter format) and enum classes (theme colors, brush cursor types, tablet driver mode probably has the same issue), saying `QVariant::save: unable to save type`.

This is fixed now and was caused by Qt5 requiring manual registration of enum types for `QDataStream`, whereas Qt6 automatically does this.

> Canvas scrollbars are always visible on startup, have to toggle them on and off to get rid of them.

This is fixed now and was caused by implicit conversion of an object pointer to bool when binding using an object for lifetime and a lambda function for the receiver. Additional template constraints have been added so objects are only considered as arguments to pointer to member functions.

askmeaboutlo0m:
> By defaulting to a non-system palette, users who require a high contrast colour scheme for accessibility reasons will find it difficult or impossible to use Drawpile because they will not be able to navigate through the preferences to switch the colour scheme back, so I would ask to please not do this. For nitpicking feedback, you can feel free to send anyone doing that in my direction if it would reduce the burden.

It's a collaborative drawing program, users who require a high contrast to the degree where they can't deal with the reasonably contrastful Krita Dark will need to apply it screen-wide anyway to actually be able to see other people's drawings, or alternatively Drawpile would need to offer a high-contrast mode that also applies to the canvas itself. So I don't think this is a valid reason to leave a bad impression with users.

askmeaboutlo0m:
I wanted to check on Windows if there was a way to query if high-contrast mode is enabled, but found that Qt, neither 5 nor 6, shows any reaction to it no matter what style or color scheme is picked, so the whole point is moot to begin with. Using the system style just gives us ugly mode, it doesn't help accessibility in any way.

But I did find a new problem: there's a windowsvista style to pick here, which has no effect and just gives the regular windows style.

image

src/desktop/dialogs/settingsdialog/tools.cpp
```diff
@@ -0,0 +1,134 @@
...
+	settings.bindShareBrushSlotColor(shareColor);
+	form->addRow(tr("Brushes:"), shareColor, 1, 2);
+
+	auto *outlineSize = new QDoubleSpinBox;
+	settings.bindBrushOutlineWidth(outlineSize);
+	outlineSize->setDecimals(1);
+	outlineSize->setMaximum(25.0);
+	outlineSize->setSingleStep(.5);
+	outlineSize->setSuffix(tr("px"));
+	auto *outlineSizeLayout = utils::encapsulate(tr("Show a %1 outline around the brush"), outlineSize);
```
askmeaboutlo0m, line 57: I'm told that this encapsulated style of setting is confusing and difficult to translate to Chinese, and thinking about it, that's the case for German as well. You end up with really long and complex sentences where before a single compound noun suffices.

askmeaboutlo0m:
Guiding a user through the tablet driver settings has proved difficult, the multiple checkboxes and radio buttons ended up just being confusing for the main use case: trying all of the options until you find one that works. So I'd like to revert it to a flat list again that the user can go down, be it a combo box or a stack of radio buttons, rather than something multi-dimensional.

PR #1184, opened 2024-02-17
Refactor the classic brush code

kiroma:
Helps during catch up and for drawing with extremely large classic brushes

### DIFF
--- a/src/drawdance/libengine/dpengine/paint.c
+++ b/src/drawdance/libengine/dpengine/paint.c
@@ -226,15 +226,15 @@ static void offset_mask(DP_BrushStamp *dst_stamp, DP_BrushStamp *src_stamp,
     uint16_t *src = src_stamp->data;
     uint16_t *dst = dst_stamp->data;
 
-    *(dst++) = DP_uint32_to_uint16((src[0] * k3) / 4);
+    *(dst++) = DP_uint32_to_uint16((src[0] * k3) / 16);
     for (int x = 0; x < diameter - 1; ++x) {
         *(dst++) =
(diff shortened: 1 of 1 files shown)


askmeaboutlo0m:
Thank you, this looks good to me! I'll give it a try.

askmeaboutlo0m:
Disregard the pipeline failures on Windows, they are not related to the code. For some reason it's trying to sign the executables when it doesn't have access to the certificate.

askmeaboutlo0m:
It's noticeably faster with large brushes, so it does what it says on the tin. The results are similar enough, there's a few single bits of difference here and there, which is fine, SIMD optimizations introduce similarly tiny differences.

But there seems to be some kind of issue with pixels at the very top and left of the stamp.

This is the expected result: old-1

This is the actual result: new-1

Here's the pixel diff between them: diff

It's particularly noticeable on the left edge of the fourth dab in the top row. The dab is black, but the pixels actually get lighter, which means they're invalid. That is, the color channels probably have higher values than the alpha channel, which shouldn't happen with premultiplied pixels. image

Here's the recording used (have to put it into a ZIP so that GitHub accepts it): [dabs.zip](https://github.com/drawpile/Drawpile/files/14320494/dabs.zip)

You can generate a PNG file from it using `drawpile-cmd -f png dabs.dprec`, which will produce a `dabs-1.png`.

If you don't have time to look into this, let me know, then I'll dig further.

kiroma:
Ah, looking now

kiroma:
Alright, the output should match now, this is what I got
dabs
Diff mask in gimp shows all black so it seems fine

askmeaboutlo0m:
Looks good now, thank you! Will merge this now.

Issue #1188, opened 2024-03-21
Flood fill: allow shrinking fills


askmeaboutlo0m:
It should contract the filled area, instead of expanding it. This should be reasonably simple to implement, since contracting the filled area just means expanding the unfilled area, so the algorithm is the same.

The following changes need to be made:

* In `flood_fill.c`:
  * `DP_flood_fill` currently clamps `expand` to a minimum of 0 when it calls `make_mask`. That needs to be removed.
  * `make_mask` should not add/subtract the `expand` value to/from the mask size when it's negative. It should just leave it as-is.
  * `make_mask` needs to use the absolute value of `expand` when generating the expansion kernel.
  * `apply_expansion_kernel` needs to invert the values when the expansion is negative. The mask uses values between 0.0 and 1.0, so inverting a value `x` means doing `1.0 - x`.
* In `fillsettings.ui`:
  * The `minimum` property of the `expand` slider needs to be set to -30. This can either be done using Qt Designer or by just editing the XML directly.
* In `fillsettings.cpp`:
  * The `expand` tool property needs to be changed to allow values from -30 to 30. Currently it goes from 0 to 100 for some reason, even though you can't select higher values than 30.

Credit Meru for suggesting in the Changelog.

askmeaboutlo0m:
Implemented slightly differently with a toggle button to switch between expanding and shrinking. Having the values go negative ended up being too annoying in practice, because centering the slider to reset it to zero is much harder than just grabbing it and slamming it all the way to the edge.

It's available in the continuous release: <https://github.com/drawpile/Drawpile/releases/tag/continuous>

PR #1190, opened 2024-03-21
fix:fix fillsettings.ui and fillsettings.cpp#1188

Dibyanshuaman:
fix:fillsettings.ui and fillsettings.cpp to allow value from -30 to 30

### DIFF
--- a/src/desktop/toolwidgets/fillsettings.cpp
+++ b/src/desktop/toolwidgets/fillsettings.cpp
@@ -13,7 +13,7 @@ namespace tools {
 
 namespace props {
 static const ToolProperties::RangedValue<int> expand{
-	QStringLiteral("expand"), 0, 0, 100},
+	QStringLiteral("expand"), -30, -30, 30},
 	featherRadius{QStringLiteral("featherRadius"), 0, 0, 40},
 	mode{QStringLiteral("mode"), 0, 0, 2},
 	size{QStringLiteral("size"), 500, 10, 9999},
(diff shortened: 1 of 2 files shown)


src/desktop/toolwidgets/fillsettings.cpp
```diff
@@ -13,7 +13,7 @@ namespace tools {
 
 namespace props {
 static const ToolProperties::RangedValue<int> expand{
-	QStringLiteral("expand"), 0, 0, 100},
+	QStringLiteral("expand"), -30, -30, 30},
```
askmeaboutlo0m, line 16: The default value should remain 0. Negative values only make sense if the rest of #1188 is actually implemented.

src/desktop/ui/fillsettings.ui
```diff
@@ -237,6 +237,9 @@
      <property name="prefix">
       <string>Expand: </string>
      </property>
+     <property name="minimum">
+       <number>-30</number>
+      </property>
```
askmeaboutlo0m, line 242: Dito to the above, negative values only make sense if the rest of #1188 is actually implemented.

askmeaboutlo0m (changes requested):
This is a partial implementation of #1188, but only the UI part of it. Negative expansion needs to actually be implemented into the flood fill for the changes to make sense.

askmeaboutlo0m:
You seem to be opening many tiny pull requests in numerous different repositories, so I don't think the issues will be addressed. Closing this PR to avoid confusing potential contributors into thinking that someone is working on this.

Feel free to reopen the PR if the issues are addressed.

PR #1202, opened 2024-04-07
Run `git_version_describe` in the source directory

fluttershydev:
By default this runs in the build directory, and if the build directory is not a subdirectory of the source directory then the command will fail.


askmeaboutlo0m:
Thanks!

Issue #1207, opened 2024-04-07
First-time setup dialog


askmeaboutlo0m:
When a new person starts using Drawpile, they usually have to do quite a bit of fiddling around in the preferences to get the program looking and feeling how they want. We should instead present them with a guided setup that does at least the following:

* Before even starting Drawpile proper, let them pick the most important parameters: their language, UI style (maybe restricting this to Bright, Dark and Darker) and maybe other stuff that normally requires restarting the application to apply.
* Maybe give them a selection of what program they're coming from so that we can set up their general settings. For example, if they come from Krita, they will almost certainly want E to toggle erase mode, not switch to a different eraser brush.
* Provide guided setup for the tablet driver, smoothing and pressure curve.
* Let you pick your initial application layout, since the "Default" layout is okay for small screens, but wastes a lot of space on large ones.

MorrowShore:
Could potentially be added to the start screen instead #1345

PR #1275, opened 2024-04-10
Change default theme to KritaDarker? @ settings_table.h

MorrowShore:
Change default theme to KritaDarker? @ settings_table.h


askmeaboutlo0m:
Krita Darker looks bad on 2 of 4 screens I have available. On those, the black is so dark that it makes it hard to look at the UI. Krita Dark looks okay on every screen, so it's safer as a default.

I think a more sensible solution is #1207, letting people pick between dark, darker and bright to start with.

Regardless of that, this implementation is not the correct way to change the default color scheme, it would also change it for everyone that currently has it set to Krita Dark.

PR #1302, opened 2024-04-27
WIP: Add pixel rulers

fluttershydev:
This adds pixel rulers to the top and left sides of the canvas. This is only implemented for the new renderer, as the previous renderer is going away eventually.

This handles auto scaling the division spacing and font size, and uses the current theme's colors for drawing the ruler.

Current issues:
* Not very pretty, would be nice to have someone mock up a better look.
* Does not support canvas rotation. Currently just turns off when the canvas is rotated.

### DIFF
--- a/src/desktop/view/canvasview.cpp
+++ b/src/desktop/view/canvasview.cpp
@@ -6,11 +6,13 @@
 #include "desktop/view/canvasscene.h"
 #include "desktop/view/glcanvas.h"
 #include "desktop/widgets/notifbar.h"
+#include "desktop/widgets/rulerwidget.h"
 #include "libclient/drawdance/eventlog.h"
 #include <QDragEnterEvent>
 #include <QDragMoveEvent>
 #include <QDropEvent>
 #include <QGestureEvent>
+#include <QGridLayout>
 #include <QMimeData>
 #include <QScopedValueRollback>
 #include <QScrollBar>
@@ -18,6 +20,8 @@
 #include <QTouchEvent>
 
 namespace view {
+
+constexpr int rulerWidth = 18;
 
 CanvasView::CanvasView(
 	CanvasController *controller, CanvasInterface *canvasWidget,
@@ -50,6 +54,38 @@ CanvasView::CanvasView(
 	connect(
 		m_notificationBar, &widgets::NotificationBar::heightChanged,
 		controller->scene(), &CanvasScene::setNotificationBarHeight);
+
+	m_hRuler = new widgets::RulerWidget(this);
+	m_hRuler->setFixedHeight(rulerWidth);
+	m_vRuler = new widgets::RulerWidget(this);
+	m_vRuler->setFixedWidth(rulerWidth);
+	m_corner = new QWidget(this);
+	m_corner->setFixedSize(rulerWidth, rulerWidth);
+	m_corner->setAutoFillBackground(true);
+	m_corner->setAttribute(Qt::WA_TransparentForMouseEvents);
+
+	auto layout = new QGridLayout(this);
+	layout->setSpacing(0);
+	layout->setContentsMargins(0, 0, 0, 0);
+	layout->addWidget(m_corner, 0, 0);
+	layout->addWidget(m_hRuler, 0, 1);
+	layout->addWidget(m_vRuler, 1, 0);
+	layout->addWidget(vp, 1, 1);
+	setLayout(layout);
+
+	m_hRuler->hide();
+	m_vRuler->hide();
+	m_corner->hide();
+
+	connect(
+		m_controller, &CanvasController::viewChanged, this,
+		&CanvasView::onViewChanged);
+	connect(
+		m_controller, &CanvasController::transformChanged, m_hRuler,
+		&widgets::RulerWidget::onTransformChanged);
+	connect(
+		m_controller, &CanvasController::transformChanged, m_vRuler,
+		&widgets::RulerWidget::onTransformChanged);
 
 	desktop::settings::Settings &settings = dpApp().settings();
 	settings.bindCanvasScrollBars(this, &CanvasView::setEnableScrollBars);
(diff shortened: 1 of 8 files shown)


fluttershydev:
What it looks like on Windows 11
image

I also profiled it on a modern Intel i9 and 97% of the time is in the QPainter drawing functions, so I didn't do anything to try to avoid recomputing the transforms, as it wouldn't really have any impact. There's 3% taken by getting the font metrics, but the overall time of the paint is so low in a release build that caching that lookup table and invalidating on font change doesn't seem worth the complexity.

fluttershydev:
Compiler errors are a bit weird. `std::array` has been `constexpr` since C++11 and had deduction guides since C++17. `std::lround` has existed since C++11. The compilers used where that's failing should support that.

src/desktop/widgets/rulerwidget.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+void RulerWidget::onViewChanged(const QPolygonF &view)
+{
+	if(isHorizontal()) {
+		m_lMin = view.boundingRect().left();
+		m_lMax = view.boundingRect().right();
+	} else {
+		m_lMin = view.boundingRect().top();
+		m_lMax = view.boundingRect().bottom();
+	}
```
askmeaboutlo0m, line 149: The widget sometimes doesn't change for me when the canvas is moved because it doesn't schedule repaints. Functions that change painting-related state need to call `QWidget::update` to let Qt know that it's painting time.

```suggestion
	}
	update();
```

src/desktop/widgets/rulerwidget.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void RulerWidget::setCanvasToRulerTransform(qreal scale, int offset)
+{
+	m_canvasToRulerScale = scale;
+	m_canvasToRulerOffset = offset;
```
askmeaboutlo0m, line 155:
```suggestion
	m_canvasToRulerOffset = offset;
	update();
```

src/desktop/widgets/rulerwidget.cpp
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void RulerWidget::onTransformChanged(qreal zoom [[maybe_unused]], qreal angle)
+{
+	m_isRotated = std::abs(angle) > 0.001;
```
askmeaboutlo0m, line 160:
```suggestion
	bool isRotated = std::abs(angle) > 0.001;
	if(isRotated != m_isRotated) {
	    m_isRotated = isRotated;
	    update();
	}
```

src/desktop/view/canvasview.cpp
```diff
@@ -19,6 +21,8 @@
 
 namespace view {
 
+constexpr int rulerWidth = 18;
```
askmeaboutlo0m, line 24: Nit: should probably be `static`. I dunno if it's still relevant, but compilers used to be very unsmart about internal linkage, but behaved properly if given static linkage.

src/desktop/widgets/rulerwidget.cpp
```diff
@@ -0,0 +1,162 @@
...
+			digitsNeeded(canvasL + adjustedPPD * divisions));
+
+		while(true) {
+			lineHeight =
+				painter.fontMetrics().tightBoundingRect("0123456789").height();
+			lineHeight += 2;
+
+			if(lineHeight * digits >= pixelsPerRulerDivision - 2) {
+				QFont smallerFont = painter.font();
+				int newPointSize = smallerFont.pointSize() - 1;
```
askmeaboutlo0m, line 102: QFont inexplicably has two incompatible ways of specifying its size, either as points or as pixels. Asking it for the "wrong" kind of size will give you -1, see <https://doc.qt.io/qt-5/qfont.html#pixelSize>.

You could handle that if you want, although other UI elements of Drawpile don't do it either. The only system where this seems to happen in practice is Android, where the font size is ridiculous to begin with and we override it with a point size by default, so the cases where you get a pixel font are probably pathological.

askmeaboutlo0m (approved):
Looks good to me! The `update` thing should probably addressed, but otherwise it's fine to merge. Let me know when you want to do so, it's fine if it's not in a complete state at that point, since it will still let folks test it.

Kinda wondering if it would make sense to wrap the canvas view into another widget and put the rulers on that instead so that they don't intrude into the canvas itself. Currently, HUD elements like the zoom notification or the small screen mode buttons don't account for the rulers taking extra room and get covered by or squashed against them. It would also escape the canvas view dichotomy and let it work for the QGraphicsView-based canvas as well.

fluttershydev:
> Looks good to me! The `update` thing should probably addressed, but otherwise it's fine to merge. Let me know when you want to do so, it's fine if it's not in a complete state at that point, since it will still let folks test it.

I'll fix the issues you pointed out first.

> Kinda wondering if it would make sense to wrap the canvas view into another widget and put the rulers on that instead so that they don't intrude into the canvas itself. Currently, HUD elements like the zoom notification or the small screen mode buttons don't account for the rulers taking extra room and get covered by or squashed against them. It would also escape the canvas view dichotomy and let it work for the QGraphicsView-based canvas as well.

I noticed the notification box issue, but I didn't even think about the small screen buttons as I don't use it in that mode. Wrapping it on the outside would probably be easier than having everything else need to be aware of the extra space. Although I'm not sure the best way to structure that between the MainWindow and CanvasWrapper.

src/desktop/view/canvasview.cpp
fluttershydev: I'm not aware of any modern compilers that mess this up. constexpr implies static here by the spec.

src/desktop/widgets/rulerwidget.cpp
fluttershydev: I'll adjust the check so it bails out in that case too. Probably better to use the default size than to try and call setPointSize with a negative number.

askmeaboutlo0m:
> Although I'm not sure the best way to structure that between the MainWindow and CanvasWrapper.

I don't have anything particular in mind for it. If you want, you can leave it be and I can have a go at it when it's merged, otherwise it should be fine to just do whatever works.

fluttershydev:
I'm fine with landing as is, I won't have time until next weekend to look into rearranging that.

askmeaboutlo0m:
Alright! Will merge this in then and see about fiddling with the widget arrangement if I get time.

PR #1343, opened 2024-07-17
Changing phrasing from "confirmation" to "details"

MorrowShore:
Changing phrasing from "confirmation" to "details". Confirmation will be confused with the preview / brush confirmation mechanism, even if in the notice category.


askmeaboutlo0m:
That's what it's referring to: the notice that shows up prompting you to confirm your previewed fill.

image

PR #1370, opened 2024-09-25
[API] Implement session and user authentication

notfood:
Authentication via the json API

- Implement `/api/session/:sessionid/auth`
- Implement `/api/session/:sessionid/auth/op`
- Implement `/api/account/auth`
- Updated documentation

### DIFF
--- a/doc/server-api.md
+++ b/doc/server-api.md
@@ -153,6 +153,23 @@ To send a message to all session participants: `PUT /api/sessions/:id/`
 
 To shut down a session: `DELETE /api/sessions/:id/`
 
+To authenticate to a session: `POST /api/sessions/:id/auth`
+
+    {
+        "password": "the password of the session"
+    }
+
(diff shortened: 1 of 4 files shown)


askmeaboutlo0m (changes requested):
It's odd that checking the session or op password is disambiguated by some argument inside of the body. A more sensible way to do this would be to have two paths, e.g. `/auth/session` to check the session password and `/auth/op` to check the operator password, for example.

In the session API, you should check the path length and `return JsonApiNotFound()` if it's an unexpected length, like other API functions do.

In both the session and the user API, you should check the request method (`JsonApiMethod::Create` for `POST`) and `return JsonApiBadMethod()` if it's incorrect.

I can merge and make those changes if you want, just lemme know.

notfood:
Made the requested changes.
- Separated password from opword
- Added size checks
- Added method checks

askmeaboutlo0m (approved):

PR #1397, opened 2024-11-02
[API] Implement session listing

notfood:
Listing via the json API

- Implements `/api/sessions/:id/listing`
- Updates documentation

### DIFF
--- a/src/libserver/session.cpp
+++ b/src/libserver/session.cpp
@@ -1673,7 +1673,7 @@ JsonApiResult Session::callJsonApi(
 		}
 		killSession(QStringLiteral("Session terminated by administrator"));
 		return JsonApiResult{
-			JsonApiResult::Ok, QJsonDocument(QJsonObject{{"status", "ok "}})};
+			JsonApiResult::Ok, QJsonDocument(QJsonObject{{"status", "ok"}})};
 	}
 
 	return JsonApiResult{
@@ -1695,7 +1695,7 @@ JsonApiResult Session::callListingsJsonApi(
 				unlistAnnouncement(a.apiUrl.toString());
 				return JsonApiResult{
 					JsonApiResult::Ok,
-					QJsonDocument(QJsonObject{{"status", "ok "}})};
+					QJsonDocument(QJsonObject{{"status", "ok"}})};
 
 			} else {
 				return JsonApiBadMethod();
--- a/src/libserver/session.cpp
+++ b/src/libserver/session.cpp
@@ -1575,19 +1575,23 @@ QJsonObject Session::getDescription(bool full) const
 				{"online", false}};
 		}
 		o["users"] = users;
-
-		QJsonArray listings;
-		const auto announcements = m_announcements->getAnnouncements(this);
-		for(const sessionlisting::Announcement &a : announcements) {
-			listings << QJsonObject{
-				{"id", a.listingId},
-				{"url", a.apiUrl.toString()},
-			};
-		}
-		o["listings"] = listings;
+		o["listings"] = getListingsDescription();
 	}
 
 	return o;
+}
+
+QJsonArray Session::getListingsDescription() const
+{
+	QJsonArray listings;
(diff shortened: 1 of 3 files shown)


src/libserver/session.h
```diff
@@ -11,6 +11,7 @@
 #include <QJsonObject>
 #include <QObject>
 #include <QString>
+#include <qjsonarray.h>
```
askmeaboutlo0m, line 14: Use `#include <QJsonArray>` instead, `<qjsonarray.h>` is not a public header.

src/libserver/session.cpp
```diff
@@ -1684,24 +1688,48 @@ JsonApiResult Session::callListingsJsonApi(
...
+		return JsonApiResult{
+			JsonApiResult::Ok, QJsonDocument(getListingsDescription())};
+			
+	} else if(method == JsonApiMethod::Create) {
+		if(request.contains("url")) {
+			makeAnnouncement(QUrl(request["url"].toString()));
+			return JsonApiResult{
+				JsonApiResult::Ok,
+				QJsonDocument(QJsonObject{{"status", "ok "}})};
+		}
```
askmeaboutlo0m, line 1728: If the URL is missing, this should return a 400 error, not fall through to a 404. Something like:

```
} else {
    return JsonApiErrorResult(JsonApiResult::BadRequest, QStringLiteral("Missing url"));
}
```

src/libserver/session.cpp
```diff
@@ -1684,24 +1688,48 @@ JsonApiResult Session::callListingsJsonApi(
...
+	if (method == JsonApiMethod::Get) {
+		return JsonApiResult{
+			JsonApiResult::Ok, QJsonDocument(getListingsDescription())};
+			
+	} else if(method == JsonApiMethod::Create) {
+		if(request.contains("url")) {
+			makeAnnouncement(QUrl(request["url"].toString()));
+			return JsonApiResult{
+				JsonApiResult::Ok,
+				QJsonDocument(QJsonObject{{"status", "ok "}})};
```
askmeaboutlo0m, line 1727: Remove extra space after `ok`.

src/libserver/session.cpp
```diff
@@ -1684,24 +1688,48 @@ JsonApiResult Session::callListingsJsonApi(
...
+		return JsonApiNotFound();
+	}
+
+	if (method == JsonApiMethod::Get) {
+		return JsonApiResult{
+			JsonApiResult::Ok, QJsonDocument(getListingsDescription())};
+			
+	} else if(method == JsonApiMethod::Create) {
+		if(request.contains("url")) {
+			makeAnnouncement(QUrl(request["url"].toString()));
```
askmeaboutlo0m, line 1724: The `makeAnnouncement` function doesn't behave ideally for this. It'll send a message into the session if listing is not allowed and always returns `void` instead of some error indicator, so the API caller won't notice the error. I think it's okay to leave it like that, since it's unlikely to cause problems in practice and it's not like the web UI gives proper error feedback anyway, but maybe add a comment about it here.

askmeaboutlo0m (changes requested):
Looks good to me, just minor bits here and there. I can also fix them myself if you want, since they're all so minor. Lemme know if you want me to do that instead.

src/libserver/session.cpp
notfood: All the responses with ok come with a space. Maybe a separate commit to remove it from all of them?

src/libserver/session.cpp
notfood: It'd need to touch more code to make it possible since the `Announcements` doesn't offer other ways. Let's leave it like that for now.

src/libserver/session.cpp
askmeaboutlo0m: Hmm, leave it like this then. It's probably a mistake, but not worth breaking things by changing the behavior.

notfood: I added a separate commit before, to fix that.

Issue #1499, opened 2025-06-20
Add pressure tester on browser start page


askmeaboutlo0m:
Basically just some kind of box that says "test your pen pressure here" and confirms that the user has pressure if it receives pen events with different pressure levels. This would allow users to check different browsers and settings without having to wait for the entire WebAssembly application to download and boot up.

This change needs to be made in [src/desktop/wasm/init.js](https://github.com/drawpile/Drawpile/blob/main/src/desktop/wasm/init.js). It doesn't involve any WebAssembly, just JavaScript and HTML, so it should be reasonably easy.

Issue #1498, opened 2025-06-20
Add language selection on browser start page


askmeaboutlo0m:
You currently can't sensibly change the language of the web browser version of Drawpile. The only way to do it is via the `locale` query parameter, which can be set to one of the [supported languages](https://github.com/drawpile/Drawpile/blob/main/cmake/DrawpileTranslations.cmake#L5), for example <https://web.drawpile.net/?locale=de_DE> for German. Note that most languages that use a non-Latin script don't work in the browser.

There should instead be a drop-down that lets the user pick the language before starting Drawpile. Each language option should be written in English and the native language, similar to how it works in the application. No flag icons! Never use flags for language selection.

It should default to the locale parameter in the URL or fall back to the browser's reported language, punting to English otherwise.

This change needs to be made in [src/desktop/wasm/init.js](https://github.com/drawpile/Drawpile/blob/main/src/desktop/wasm/init.js). It doesn't involve any WebAssembly, just JavaScript and HTML, so it should be reasonably easy.

To be able to test this properly, it's still kinda necessary [to build the WebAssembly version of Drawpile](https://docs.drawpile.net/help/development/buildingfromsource#webassembly), which is a bit involved, so it probably makes most sense for someone to work on it that also wants to play around with WebAssembly. If not, itt's also okay to write it "blindly" on a separate page and just submit a PR to be tested. It increases the length of feedback cycles a bit, but probably wouldn't be too bad with a small feature like this.

PR #1501, opened 2025-06-20
Added pen pressure tester to browser startup page

shivt-F5:
Added pen pressure tester to browser startup page as suggested (https://github.com/drawpile/Drawpile/issues/1499#issue-3163232083)


askmeaboutlo0m:
This looks pretty good! The only relevant issue I can see is that it jumps back from "pen pressure detected!" to "detected pen input but no pressure" once the pen is lifted. Pretty sure this is just because it still reacts to the pen inputs even when the pen is not down, otherwise it seems to work well.

And, while the canvas to show your pen inputs is cute, I probably wouldn't add this here. It's just too likely that it will confuse artists when they see something like this in reaction to fast pen strokes, with the dots being disconnected:

Pen tester dots spread apart

And then they will think something is wrong with their tablet and try to fix it or ask for support about it. That could be fixed by converting the dots into a spline and draw a pretty curve out of them, but that's really far too much for this purpose. It's enough if it says that pressure was detected successfully or not.

And oh right, um, I guess it doesn't work out of the box because it expects a qtloader.js to be laying around. You could copy the files off of <https://web.drawpile.net/> and replace the init.js file I guess, but it's not necessary, just removing the reaction to a hovering pen and removing the canvas will be enough to integrate it. I'll keep it in mind for #1498 though, since that definitely needs instructions on how to grab the pages for testing.

shivt-F5:
hope this is what you were looking for! sorry for the extra bits from before, I got rid of the canvas and edited the pressure message to lock once it pops up once.

askmeaboutlo0m:
No need to apologize! If you want to play with it yourself, figuring out how to make splines from those points is somewhat interesting and likely come in handy knowing.

This looks good anyway, thank you! I'll integrate it and maybe fiddle a bit with the positioning and such, it's not worth it for you to build the entire WASM application just to get that page working properly. Thanks for contributing!

PR #1503, opened 2025-06-22
OKLAB simd variants draft

GloopShlugger:
I believe everything is implemented, it just needs some cleanup:
- there's some inconsistent naming from copying other functions as templates
- I think a single function accepts r g b a instead of b g r a 

The lasso tool doesn't write the final output when using either of the SIMD variants, it does however work in recolor mode which is a little odd


askmeaboutlo0m:
Effectively merged with #1512

Issue #1502, opened 2025-06-20
Allow save and reconnect on Android and browser


askmeaboutlo0m:
Since they don't support opening multiple windows, Android and browser will ask the user whether they want to discard their image when reconnecting to a session. If they want to save their image first, they have to close that dialog, go to save the image themselves and the repeat the reconnect attempt.

The application should just give them the option to save and reconnect to get rid of this rigmarole. It should wait until the saving is done and then proceed, stopping the process if saving fails or is cancelled.

The same could be done for other cases where the user is prompted like this, like when they want to open a file. Probably makes sense, since it's the same code path.

Credit tiar with suggesting.

askmeaboutlo0m:
Implemented along with fast reconnects.

PR #1510, opened 2025-07-05
Issue 1502

shivt-F5:
working on resolving save and reconnect issue: https://github.com/drawpile/Drawpile/issues/1502


shivt-F5:
hi, just a follow up regarding this

askmeaboutlo0m:
I don't think it's quite that easy. Saving is asynchronous, so what this will do is start a save, then obliterate the window it belongs to and if you didn't have a file chosen yet it won't actually save. It needs to actually wait for the save to complete before continuing.

askmeaboutlo0m:
#1502 is now resolved.

Issue #1389, opened 2024-10-15
Canvas view zoom out aliasing


cromachina:
Drawpile version 2.2.1, Linux

### Problem
When looking at a large canvas while zoomed out (for example, 4000x4000 on a 1080p screen), the canvas view looks extremely aliased, as if using nearest neighbor downscaling. I checked in the `Preferences -> General -> Canvas View: Interpolate when view is zoomed or rotated`, and it only seems to apply when zoomed in, not zoomed out.

### Potential solution?
I'm not quite sure how it works in Drawpile, but I'm guessing it's because of the limited resampling options provided for `QPixmaps`, `Qt::SmoothTransformation` (bilinear, probably) and `Qt::FastTransformation` (nearest neighbor), both of which tend to look pretty bad when downsampling large images. For my own paint tool project, I hacked around this by using `OpenCV::resize` to downscale the composited canvas view with `INTER_AREA` interpolation: https://github.com/cromachina/crowpainter/blob/main/src/crowpainter/main.py#L269

`OpenCV::resize` seems to be really well optimized and parallelized, but you could also compute a zoom level pyramid or mipmap off of the main thread and update the view later, async (kind of like how Paint Tool SAI does it), if there are performance or library usage concerns.

Aliased canvas example in Drawpile:
![image](https://github.com/user-attachments/assets/c2acb485-9024-4d74-959a-89700ce84755)

Smooth canvas example using my own tool:
![image](https://github.com/user-attachments/assets/7231ae36-5064-4301-86dd-2de4b1bd5d47)

askmeaboutlo0m:
The behavior is correct, the analysis not quite. It's not using nearest-neighbor interpolation, you just run into the limits of bilinear interpolation at those kinds of zooms and end up with not very smooth results.

There's currently three different canvas renderers: one based on QGraphicsView that'll hopefully get thrown out at some point, a "software" renderer using QPainter directly and a hardware renderer that uses OpenGL (on Windows, it's Direct3D under the hood via ANGLE.) They all work a bit differently.

In the first two cases, there's [the QGraphicsView CanvasItem paint function](https://github.com/drawpile/Drawpile/blob/main/src/desktop/scene/canvasitem.cpp#L73) and [the software renderer paint function](https://github.com/drawpile/Drawpile/blob/main/src/desktop/view/softwarecanvas.cpp#L91). Since you can't control how QPainter does its thing, you'd probably have to generate LOD instances of the canvas to feed to it.

[...]

PR #1527, opened 2025-10-06
Feature/improve hardware zoom quality

cromachina:
Addresses https://github.com/drawpile/Drawpile/issues/1389

Tested on Linux.
Tested combinations of drawing, zooming, and switching interpolation modes.

Example of a zoomed out viewport with a 4000x4000 file, with the hardware renderer (view screenshots at full res):
With the interpolate setting on:
image

With the interpolate setting off:
image


cromachina:
Android fails because the symbol `GL_TEXTURE_LOD_BIAS` is missing. I'm not sure how it's defined on Android.

askmeaboutlo0m:
Looks good, thanks for contributing!

PR #1535, opened 2025-10-17
Haiku build fixes

Begasus:
Define Haiku as __HAIKU__ and Haiku doesn't use -ldl


src/desktop/bundled/miniaudio/CMakeLists.txt
```diff
@@ -15,6 +15,8 @@ target_compile_definitions(miniaudio PUBLIC
 # Some platforms need extra flags.
 if(APPLE)
     target_link_options(miniaudio PUBLIC -lpthread -lm)
+elseif(HAIKU)
+    target_link_libraries(miniaudio PUBLIC -latomic -lpthread -lm)
```
Begasus: Checking if add "AND NOT HAIKU" in the next statement fixes the build also, if that works thise "elseif()" will be dropped.

PR #1569, opened 2026-05-14
Windows thumbnail shell extension

cow54787:
## Description

Adds `drawpile_shell_ext.dll` on windows. This is a shell extension that provides thumbnails for `.dpcs` files.   
The WIX installer has been updated to register the DLL as a shell extension.  
The code is self contained, only relying on windows libraries (WRL for COM) and sqlite3. WIC (Windows Imaging Component) is used to parse the JPEG thumbnail inside the `.dpcs` file.

The PR also contains a simple test exe that loads the provider via COM and outputs a single thumbnail. I used it to test the shell extension and use a debugger, without having to rely on the file explorer.

The extension can be expanded in the future to support more file formats or other windows integration.

## Formalities

I did use LLMs to understand various windows APIs.

### DIFF
--- a/cmake/DrawpileOptions.cmake
+++ b/cmake/DrawpileOptions.cmake
@@ -53,7 +53,7 @@ if(NOT ANDROID AND NOT EMSCRIPTEN)
 	add_feature_info("Benchmarks (BENCHMARKS)" BENCHMARKS "")
 
 	cmake_dependent_option(
-		WIN_SHELL_EXT "Build Windows shell extension for .dpcs thumbnail provider"
+		WIN_SHELL_EXT "Build Windows shell extension for .dpcs/.dppr thumbnail provider"
 		ON "WIN32;CLIENT" OFF)
 	add_feature_info("Windows shell extension (WIN_SHELL_EXT)" WIN_SHELL_EXT "")
 else()
--- a/cmake/DrawpileOptions.cmake
+++ b/cmake/DrawpileOptions.cmake
@@ -51,6 +51,11 @@ if(NOT ANDROID AND NOT EMSCRIPTEN)
 
 	option(BENCHMARKS "Build benchmarks" OFF)
 	add_feature_info("Benchmarks (BENCHMARKS)" BENCHMARKS "")
+
+	cmake_dependent_option(
+		WIN_SHELL_EXT "Build Windows shell extension for .dpcs thumbnail provider"
+		ON "WIN32;CLIENT" OFF)
+	add_feature_info("Windows shell extension (WIN_SHELL_EXT)" WIN_SHELL_EXT "")
 else()
 	# CMake allows unexposed options to be enabled
 	set(SERVER OFF CACHE BOOL "" FORCE)
--- a/src/win-shell-ext/ThumbnailProvider.cpp
+++ b/src/win-shell-ext/ThumbnailProvider.cpp
@@ -1,3 +1,5 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
 #include "ThumbnailProvider.h"
 #include "dpcs.h"
 #include <wrl/module.h>
@@ -44,7 +46,12 @@ HRESULT STDMETHODCALLTYPE ThumbnailProvider::GetThumbnail(
 		return hr;
 	}
 
-	sqlite3_close(db);
+	int rc = sqlite3_close(db);
+	if(rc != SQLITE_OK) {
+		fprintf(stderr, "Fail to close database: %s\n", sqlite3_errmsg(db));
+
+		return E_FAIL;
+	}
 
 	return S_OK;
 }
--- a/src/win-shell-ext/ThumbnailProvider.cpp
+++ b/src/win-shell-ext/ThumbnailProvider.cpp
@@ -1,0 +1,52 @@
+#include "ThumbnailProvider.h"
+#include "dpcs.h"
+#include <wrl/module.h>
+
+using namespace Microsoft::WRL;
+
+HRESULT STDMETHODCALLTYPE ThumbnailProvider::Initialize(IStream *pstream, DWORD)
+{
+	if(!pstream)
+		return E_INVALIDARG;
+	if(m_pStream)
+		return HRESULT_FROM_WIN32(ERROR_ALREADY_INITIALIZED);
+	m_pStream = pstream;
+
+	return S_OK;
+}
+
+HRESULT STDMETHODCALLTYPE ThumbnailProvider::GetThumbnail(
+	UINT cx, HBITMAP *phbmp, WTS_ALPHATYPE *pdwAlpha)
+{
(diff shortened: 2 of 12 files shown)


cmake/DrawpileOptions.cmake
```diff
@@ -51,6 +51,11 @@ if(NOT ANDROID AND NOT EMSCRIPTEN)
 
 	option(BENCHMARKS "Build benchmarks" OFF)
 	add_feature_info("Benchmarks (BENCHMARKS)" BENCHMARKS "")
+
+	cmake_dependent_option(
+		WIN_SHELL_EXT "Build Windows shell extension for .dpcs thumbnail provider"
```
askmeaboutlo0m: Not important: I guess ".dpcs/.dppr" would be slightly more accurate.

src/win-shell-ext/dpcs.cpp
```diff
@@ -0,0 +1,182 @@
...
+	return S_OK;
+}
+
+HRESULT OpenSqlite(std::vector<unsigned char> &databaseBuffer, sqlite3 **out)
+{
+	int rc;
+	rc = sqlite3_open(":memory:", out);
+	if(rc != SQLITE_OK) {
+		fprintf(stderr, "Cannot open database: %s\n", sqlite3_errmsg(*out));
+		sqlite3_close(*out);
```
askmeaboutlo0m, line 40: I was slightly suspicious of getting an error message and closing a pointer that is potentially null, but SQLite does the right thing here: the error you get is a malloc failure and the closure just returns harmlessly. So this is all good.

src/win-shell-ext/dpcs.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+	return S_OK;
+}
+
+HRESULT ExtractJPEG(sqlite3 *db, std::vector<unsigned char> &out)
+{
+	sqlite3_stmt *stmt;
+	int rc;
+	rc = sqlite3_prepare_v2(
+		db, "SELECT thumbnail FROM snapshots ORDER BY taken_at DESC LIMIT 1",
```
askmeaboutlo0m: Better to `ORDER BY snapshot_id DESC` here. The id uses `AUTOINCREMENT`, which means it will always increase and not re-use past ids. Ordering by `taken_at` could give the wrong thumbnail if the user has an incorrect system time set.

src/win-shell-ext/dpcs.cpp
```diff
@@ -0,0 +1,182 @@
...
+	if(FAILED(hr))
+		return hr;
+
+	ComPtr<IWICStream> wicStream;
+	if(FAILED(hr = factory->CreateStream(&wicStream)))
+		return hr;
+	// InitializeFromMemory takes non-const BYTE* but does not modify the buffer
+	if(FAILED(
+		   hr = wicStream->InitializeFromMemory(
+			   data, static_cast<int32_t>(size))))
```
askmeaboutlo0m: Not important, but if you want to type less, you can use `int32_t(size)` for stuff like this. It uses the conversion constructor, which is functionally identical to a static cast, just less wordy.

Of course that only works for types whose names don't contain a space, for `long long` or something this doesn't work because C++ is a fun language.

src/win-shell-ext/dpcs.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+	uint32_t srcW = 0, srcH = 0;
+	if(FAILED(hr = frame->GetSize(&srcW, &srcH)))
+		return hr;
+
+	// Scale to fit within cx×cx, maintaining aspect ratio
+	uint32_t outW, outH;
+	if(srcW >= srcH) {
+		outW = cx;
+		outH = std::max(1u, srcH * cx / srcW);
```
askmeaboutlo0m: Shouldn't these take a detour via `double` for precision? Or at least `unsigned long long`/`uint64_t` to avoid overflow.  Then again, the result is evidently correct and the thumbnails won't be so ginormous in practice that they'll overflow, so it probably doesn't matter.

src/win-shell-ext/main.cpp
```diff
@@ -0,0 +1,111 @@
...
+	L"Software\\Classes\\CLSID\\{3C98D577-6E3D-4C7F-B4A2-1D9F82E5C301}";
+static constexpr wchar_t kKeyInproc[] =
+	L"Software\\Classes\\CLSID\\{3C98D577-6E3D-4C7F-B4A2-1D9F82E5C301}"
+	L"\\InprocServer32";
+static constexpr wchar_t kKeyShellExDpcs[] =
+	L"Software\\Classes\\.dpcs\\ShellEx\\{E357FCCD-A995-4576-B01F-"
+	L"234630154E96}";
+static constexpr wchar_t kKeyShellExDppr[] =
+	L"Software\\Classes\\.dppr\\ShellEx\\{E357FCCD-A995-4576-B01F-"
+	L"234630154E96}";
```
askmeaboutlo0m, line 72: Assuming that these GUIDs are freshly generated or predefined and not taken from a sample. If not: remember to also change them in the CLSID and the WIX template.

src/win-shell-ext/ThumbnailProvider.cpp
```diff
@@ -0,0 +1,52 @@
...
+		sqlite3_close(db);
+		return hr;
+	}
+
+	if(FAILED(hr = JpegToHBitmap(jpegBuf.data(), jpegBuf.size(), cx, phbmp))) {
+		sqlite3_close(db);
+		return hr;
+	}
+
+	sqlite3_close(db);
```
askmeaboutlo0m: Closing the database can fail if statements are left open, so it would be slightly nicer to wrap this and print a warning. Although it's also fine to leave it this way, since it is only something relevant when developing and testing, nobody's going to see the thumbnailer printfs in practice.

askmeaboutlo0m (commented):
Looks good, only the `ORDER BY snapshot_id DESC` is functionally relevant, the rest is whatever.

`SPDX-License-Identifier` commands would also be nice to make it explicit. Just `// SPDX-License-Identifier: GPL-3.0-or-later` or `// SPDX-License-Identifier: MIT` or some other compatible thing at the top is fine. The Microsoft thumbnailer samples are MIT-licensed.

src/win-shell-ext/main.cpp
cow54787: The UUID for the provider `3C98D577-6E3D-4C7F-B4A2-1D9F82E5C301` has been freshly generated.   
`E357FCCD-A995-4576-B01F-234630154E96` is a windows specific UUID for thumbnail providers extension: https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/cc144118(v=vs.85)#registering-your-thumbnail-handler
> The GUID for the thumbnail handler Shell extension is E357FCCD-A995-4576-B01F-234630154E96.

PR #1575, opened 2026-05-30
Use native image fallback for file image imports

### LINKED ISSUE #1574: WebP files may fail when dropped or pasted as image data
Dragging or pasting a WEBP file into the canvas or the reference dock may fail to open the file, whereas opening it via File > Open works fine. Similary, QOI works only with the latter.

This is caused by Drawpile using different code paths to open these, one using Qt's image loader and the other using the Drawdance one. They should both use the same path for consistency.

LynnColeArt:
## Description

Fixes #1574.

This makes file-based image imports use a shared loader that tries both Qt image loading and Drawdance's native image importer. WebP and QOI prefer the native importer first, then fall back to Qt, while other formats keep the Qt-first behavior.

It also makes canvas drag/drop prefer local file URLs before decoded image MIME data, so dropped files can use the same file import path as manual opens instead of depending on the desktop environment's decoded image payload.

Covered entry points include:

- opening image file content
- pasting a file path
- canvas file drops
- reference image drops

## Formalities

The code is new. It was developed in a GPLv3 fork of Drawpile and is being contributed back under Drawpile's existing license. No code was copied from another project.

AI assistance was used to help identify the inconsistent import paths, draft the initial implementation, and isolate the patch into this upstream-focused branch. The changes were reviewed locally before submission.

## Testing

- `git diff --check upstream/main...HEAD`
- `ninja -C build-qt5-client-baseline drawpile`

### DIFF
--- a/src/desktop/docks/reference.cpp
+++ b/src/desktop/docks/reference.cpp
@@ -7,10 +7,10 @@
 #include "desktop/widgets/kis_slider_spin_box.h"
 #include "desktop/widgets/referenceview.h"
 #include "libclient/document.h"
+#include "libclient/utils/images.h"
 #include <QAction>
 #include <QButtonGroup>
 #include <QHBoxLayout>
-#include <QImageReader>
 #include <QMenu>
 #include <QScrollBar>
 #include <QSignalBlocker>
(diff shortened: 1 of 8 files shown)


askmeaboutlo0m:
AI-generated code has licensing and obvious ethical problems. It will not be accepted and I will not waste time reviewing it.

LynnColeArt:
Good to know. 
Thank you. 

But if it's banned you should say so in CONTRIBUTORS, or the README, or ideally both.

I was trying to be conscientious of this exact thing, and found no hard warning signs that said it couldn't be done.

Asking in the PR template isn't enough. You need to state it clearly.

You could also put up a CLAUDE or AGENTS file stipulates the decision.

Be more explicit in your concerns, and people will stop doing it.

Nobody wants to contribute to projects where their method of contribution is unwanted.

You probably won't hear from me again.

Take care.

PR #1577, opened 2026-06-09
fix: load WebP/QOI via Drawdance on drop and paste

### LINKED ISSUE #1574: WebP files may fail when dropped or pasted as image data

mvanhorn:
## Description

Dropping or pasting a WebP file (and likewise QOI) could fail to open even though `File > Open` loads the same file fine. The two paths used different decoders: `File > Open` goes through the Drawdance impex layer (`DP_image_new_from_file`), which supports WebP/QOI, while drop/paste/reference went through Qt's loader (`QImageReader` / `QImage(path)` / `QImage::loadFromData`), which lacks those decoders. A member confirmed this root cause on the issue.

This consolidates the loading so dropped and pasted images support the same formats as opened files. Fixes #1574.

Changes:

- `src/libclient/drawdance/image.{h,cpp}`: add `drawdance::loadImage` helpers (one for a file path, one for an in-memory `QByteArray`) that decode through the Drawdance impex loader and fall back to Qt's loader only when Drawdance returns nothing, so formats Qt already handles are unaffected.
- `src/desktop/docks/reference.cpp`: route the reference-dock file drop through `drawdance::loadImage`.
- `src/libclient/document.cpp`: `getClipboardImageData` keeps the native in-memory `QImage` fast path, then decodes raw `image/*` clipboard byte buffers (e.g. `image/png`, `image/webp`) through `drawdance::loadImage`. `clipboardHasImageData` now also reports image data when the clipboard advertises any raw `image/*` format, so the Paste action is enabled for WebP/QOI.
- `src/desktop/mainwindow.cpp`: route the dropped-file paste path (`pasteFilePath`) through `drawdance::loadImage`.
- `src/desktop/widgets/referenceview.cpp`: accept reference-dock file drops for the Drawdance flat image formats (`cmake_config::file_group::flatImage()`: PNG/JPEG/QOI/WEBP) in addition to Qt-supported formats, so QOI (and WebP on builds without Qt's WebP plugin) is no longer rejected before `handlePathDrop` runs.

Native in-memory `QImage` clipboard data keeps its existing fast path; only raw byte buffers and file paths gain the Drawdance fallback.

Testing: I could not run a full Qt+Rust build locally. The changes were verified for syntactic consistency, helper-signature/call-site wiring, and Drawdance API correctness (`DP_file_input_new_from_path`, `DP_mem_input_new_keep_on_close`, `DP_input_free`, `DP_image_new_from_file`, and `drawdance::wrapImage` ownership). Suggested manual checks: drop/paste a `.webp` and `.qoi` onto the canvas and reference dock (should open), and confirm PNG/JPEG plus native in-memory `QImage` clipboard data still work as before.

## Formalities

The code is new. It reuses Drawpile's existing Drawdance impex loader (`DP_image_new_from_file`) and the existing `drawdance::wrapImage` helper, both already in this repository; no external code was brought in.

AI assistance was used to help draft the decode helper and route the three call sites, followed by manual review of the Drawdance API usage, memory ownership, and the format-gating logic.

Fixes #1574

### DIFF
--- a/src/desktop/docks/reference.cpp
+++ b/src/desktop/docks/reference.cpp
@@ -6,11 +6,11 @@
 #include "desktop/widgets/groupedtoolbutton.h"
 #include "desktop/widgets/kis_slider_spin_box.h"
 #include "desktop/widgets/referenceview.h"
+#include "libclient/drawdance/image.h"
 #include "libclient/document.h"
 #include <QAction>
 #include <QButtonGroup>
 #include <QHBoxLayout>
-#include <QImageReader>
 #include <QMenu>
 #include <QScrollBar>
 #include <QSignalBlocker>
@@ -237,14 +237,8 @@ void ReferenceDock::handleImageDrop(const QImage &img)
 
 void ReferenceDock::handlePathDrop(const QString &path)
 {
-	QImage img;
 	QString error;
-	{
-		QImageReader reader(path);
-		if(!reader.read(&img)) {
-			error = reader.errorString();
-		}
-	}
+	QImage img = drawdance::loadImage(path, &error);
 	showReferenceImageFromFile(img, error);
 }
 
--- a/src/desktop/mainwindow.cpp
+++ b/src/desktop/mainwindow.cpp
@@ -85,6 +85,7 @@ extern "C" {
 #include "libclient/config/config.h"
 #include "libclient/document.h"
 #include "libclient/drawdance/eventlog.h"
+#include "libclient/drawdance/image.h"
 #include "libclient/drawdance/perf.h"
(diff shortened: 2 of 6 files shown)


askmeaboutlo0m:
This is slop, that you reviewed this manually is a lie. Blocked and reported for spam.

PR #1584, opened 2026-07-18
Fix JPEG export in drawpile-cmd on Windows

### LINKED ISSUE #1582: Drawpile-cmd does not generate JPG/JPEGs on Windows
When running one of the commands below:

```
drawpile-cmd.exe -f jpg -i "D:\Users\Username\Downloads\2026-06-21 - Some File.ora"
drawpile-cmd.exe -f dpcs -i "D:\Users\Username\Downloads\2026-06-21 - Some File.ora"
```
One of two error codes appear as a result for each respective line.
```
Error 2 writing JPEG image: Unsupported image format
[WARN] src\drawdance\libengine\dpengine\snapshots.c:486 - Failed to generate reset thumbnail: Error 2 writing JPEG image: Unsupported image format
```

Apparently, this is a problem limited only to Windows, as this tool works properly on other platforms.

floze-the-genius:
## Description

`drawpile-cmd` uses Qt image I/O when it is built alongside the client or server. On Windows, JPEG support is provided by the deployed `qjpeg` plugin, but the command did not create a `QCoreApplication`, so Qt did not initialize its application/plugin search paths and JPEG export failed with `Unsupported image format`.

This changes the command entry point to C++, keeps a `QCoreApplication` alive for the command lifetime, and adds an end-to-end CTest that converts the existing PNG fixture to JPEG and verifies the `FF D8 FF` signature.

Fixes #1582.

## Validation

- `cargo test --package drawpilecmd`
- `cargo clippy --package drawpilecmd -- ...` (completed with pre-existing warnings)
- `cargo doc --package drawpilecmd --no-deps`
- Clang syntax check for the new C++ entry point
- Linked and ran the entry point against Qt with a stub asserting that `QCoreApplication::instance()` is available
- Ran the new CMake regression script with a real JPEG fixture
- `git diff --check`

A full local CMake link was not available with the minimal tools-only configuration because this repository only adds the native `drawdance_client` target when the client or server is enabled. The upstream Windows tools matrix is the authoritative end-to-end build and regression gate. `cargo fmt --all -- --check` also reports a pre-existing ordering difference in `src/drawdance/rust/msg/mod.rs`; this PR does not modify Rust sources.

## Formalities

The code is new and uses existing Qt and Drawpile APIs already present in this repository.

OpenAI Codex was used to investigate the issue, draft the entry-point change and regression test, and run validation. I reviewed the resulting diff and validation output before submitting.

### DIFF
--- a/src/tools/CMakeLists.txt
+++ b/src/tools/CMakeLists.txt
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: MIT
 
 set_property(SOURCE dprectool/main.c PROPERTY SKIP_AUTOGEN ON)
-set_property(SOURCE drawpile-cmd/main.c PROPERTY SKIP_AUTOGEN ON)
+set_property(SOURCE drawpile-cmd/main.cpp PROPERTY SKIP_AUTOGEN ON)
 
 add_cargo_library(dprectool_rust dprectool)
 add_executable(dprectool dprectool/main.c)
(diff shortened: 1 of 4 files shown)


askmeaboutlo0m:
AI-generated pull requests are not looked at.

